Issue #2, opened 2018-10-17
POM does not satisfy the SOPs


chahuistle:
This project, too, needs to be migrated to use parent POMs, etc.

PR #4, opened 2018-10-29
Sync lib repo with current template specification

sven1103:
Solves #2


sven1103:
Not sure about the license though, were third party dependencies unter GPL3?

sven1103:
@chahuistle @wow-such-code could anyone review and comment? Thanks!

PR #7, opened 2018-11-06
Feature/openbis requester

sven1103:
First example for as submitter interface, that will accept openBIS query closures and automatically converts the returned objects based on a dictionary of known adapters.


sven1103:
outdated

PR #27, opened 2020-07-06
Update parent-pom version to 2.3.0 since it includes gapidoc generation and reformatted pom.xml

Steffengreiner:


sven1103:
Merged it, as the drop of code coverage was not due to new code.

PR #30, opened 2020-07-10
Provide access to the measurement adapter

sven1103:
This commit allows the client to access the measurement
adapter used in the Oxford Nanopore Measurement.

As the adapter field is an optional metadata property, it will
be an empty String if the field was not set.


wow-such-code (dismissed):
looking good

src/test/groovy/life/qbic/datamodel/datasets/datastructure/OxfordNanoporeMeasurementSpec.groovy
```diff
@@ -193,4 +193,32 @@ class OxfordNanoporeMeasurementSpec extends Specification {
 
     }
 
+    def "missing adapter metadata should be allowed but return empty String"() {
```
KochTobi, line 196:
```suggestion
    def "empty String is returned for missing adapter metadata"() {
```

KochTobi (dismissed):
I would rename the test.

KochTobi (approved):
looks good

PR #32, opened 2020-07-16
Reflect illegal states of ON measurements

### LINKED ISSUE #31: Nanopore Instrument Data does not allow empty fast*_fail folders
https://github.com/qbicsoftware/data-model-lib/blob/master/src/main/resources/schemas/nanopore-instrument-output.schema.json#L228

The `fastq_fail`, `fastq_pass`, `fast5_fail`, `fast5_pass` folder might be empty if no Phred score is too low or high enough.

sven1103:
Fix #31 

As a failsafe and part of our business rule at QBiC,
we must ensure that the pass and fail folder for either
fast5 or fastq are not empty during the creation of
an OxfordNanoporeMeasurement object.

Therefore, if the pass and fail folder are both empty for either
fastq oder fast5, an IllegalStateException will be thrown
during the instanciation of an OxfordNanopreMeasurement object.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
 		<version>2.2.0</version>
 	</parent>
 	<artifactId>data-model-lib</artifactId>
-	<version>1.9.0-SNAPSHOT</version>
+	<version>1.10.0-SNAPSHOT</version>
 	<packaging>jar</packaging>
 	<name>Data Model Library</name>
 	<url>http://github.com/qbicsoftware/data-model-lib</url>
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
@@ -57,6 +57,7 @@ final class OxfordNanoporeMeasurement {
         readMetaData(metadata)
         createContent()
         assessPooledStatus()
+        assessState()
     }
 
     private List<DataFile> getLogFileCollection() {
@@ -78,11 +79,13 @@ final class OxfordNanoporeMeasurement {
     }
 
     private void assessPooledStatus() {
-        this.pooledSamplesMeasurement = folders["fast5pass"] ? folders["fast5pass"].getChildren().get(0) instanceof Fast5Folder : false
+        this.pooledSamplesMeasurement = folders["fast5pass"]?.getChildren() ?
+            folders["fast5pass"].getChildren().get(0) instanceof Fast5Folder : false
         // There can be still pooled samples in the failed folder, worst case is all
         // samples failed, so we need to check there to
         if (! pooledSamplesMeasurement) {
-            this.pooledSamplesMeasurement = folders["fast5fail"] ? folders["fast5fail"].getChildren().get(0) instanceof Fast5Folder : false
+            this.pooledSamplesMeasurement = folders["fast5fail"]?.getChildren() ? folders["fast5fail"]
+                .getChildren().get(0) instanceof Fast5Folder : false
         }
     }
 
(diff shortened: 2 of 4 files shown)


src/test/groovy/life/qbic/datamodel/datasets/datastructure/OxfordNanoporeMeasurementSpec.groovy
```diff
@@ -148,4 +148,41 @@ class OxfordNanoporeMeasurementSpec extends Specification {
 
     }
 
+
+    def "If both fast5 pass and fail folder are empty, an IllegalStateException shall be thrown"() {
```
wow-such-code, line 152: I think at least one test case with one folder existing and the other missing should be added

pom.xml
```diff
@@ -8,7 +8,7 @@
 		<version>2.2.0</version>
 	</parent>
 	<artifactId>data-model-lib</artifactId>
-	<version>1.9.0-SNAPSHOT</version>
+	<version>1.10.0-SNAPSHOT</version>
```
KochTobi, line 11: I think this is only a small bug fix and not a whole feature? So 
```suggestion
	<version>1.9.1-SNAPSHOT</version>
```

KochTobi (dismissed):
Looks good

pom.xml
sven1103: jeah, but it introduces an Exception, that might lead to unexpected behaviour in existing implementations. According to semantic versioning i was not 100% sure if a patch release is correct.

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -78,11 +79,13 @@ final class OxfordNanoporeMeasurement {
     }
 
     private void assessPooledStatus() {
```
wow-such-code, line 81: If nested if/nullsafe cases must be used, please add informative comments

src/main/resources/schemas/nanopore-instrument-output.schema.json
```diff
@@ -24,7 +24,7 @@
           "minLength": 1
         },
         "children": {
-          "description": "Describes one or more files and/or sub-folders",
+          "description": "Describes files and/or sub-folders if existant",
```
wow-such-code, line 27: existent*

KochTobi:
```suggestion
          "description": "Describes files and/or sub-folders if existent",
```

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
sven1103: i did not touch this line of code in this PR, nore was it part of this fix PR.

src/main/resources/schemas/nanopore-instrument-output.schema.json
sven1103: you can make a suggestion, that ca be merged so you get contribution

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -78,11 +79,13 @@ final class OxfordNanoporeMeasurement {
...
-        this.pooledSamplesMeasurement = folders["fast5pass"] ? folders["fast5pass"].getChildren().get(0) instanceof Fast5Folder : false
+        this.pooledSamplesMeasurement = folders["fast5pass"]?.getChildren() ?
+            folders["fast5pass"].getChildren().get(0) instanceof Fast5Folder : false
         // There can be still pooled samples in the failed folder, worst case is all
         // samples failed, so we need to check there to
         if (! pooledSamplesMeasurement) {
-            this.pooledSamplesMeasurement = folders["fast5fail"] ? folders["fast5fail"].getChildren().get(0) instanceof Fast5Folder : false
+            this.pooledSamplesMeasurement = folders["fast5fail"]?.getChildren() ? folders["fast5fail"]
+                .getChildren().get(0) instanceof Fast5Folder : false
```
wow-such-code, line 88: This is a bit too crazy for me without comments :)

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
wow-such-code: sorry, meant the following lines, see new comment

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
sven1103: sure, i will make it more declarative

wow-such-code (changes requested):
some small things

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
KochTobi: I think this is fine. It first checks if there are any children of the fast5fail and then assigns the first child as a Fast5Folder. In case there are no children of the fast5fail it assigns false

wow-such-code: I think the fact that you missed the first check:
`folders["fast5fail"]?`
shows that it isn't that easy to read. This is giving me Functional Programming PTSD.

sven1103:
Decided to make it as a patch

pom.xml
KochTobi: It should not. Existing implementations always have at least one item and are thus not affected.

PR #33, opened 2020-07-17
Reflect illegal states of ON measurements

### LINKED ISSUE #31: Nanopore Instrument Data does not allow empty fast*_fail folders

sven1103:
Fix #31 

As a failsafe and part of our business rule at QBiC,
we must ensure that the pass and fail folder for either
fast5 or fastq are not empty during the creation of
an OxfordNanoporeMeasurement object.

Therefore, if the pass and fail folder are both empty for either
fastq oder fast5, an IllegalStateException will be thrown
during the instanciation of an OxfordNanopreMeasurement object.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
@@ -79,14 +79,21 @@ final class OxfordNanoporeMeasurement {
     }
 
     private void assessPooledStatus() {
-        this.pooledSamplesMeasurement = folders["fast5pass"]?.getChildren() ?
-            folders["fast5pass"].getChildren().get(0) instanceof Fast5Folder : false
+        this.pooledSamplesMeasurement = containsAtLeastOneBarcodedFolder(folders["fast5pass"])
         // There can be still pooled samples in the failed folder, worst case is all
         // samples failed, so we need to check there to
         if (! pooledSamplesMeasurement) {
-            this.pooledSamplesMeasurement = folders["fast5fail"]?.getChildren() ? folders["fast5fail"]
-                .getChildren().get(0) instanceof Fast5Folder : false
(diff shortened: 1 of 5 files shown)


src/main/resources/schemas/nanopore-instrument-output.schema.json
```diff
@@ -24,7 +24,7 @@
           "minLength": 1
         },
         "children": {
-          "description": "Describes one or more files and/or sub-folders",
+          "description": "Describes files and/or sub-folders if existant",
```
wow-such-code, line 27:
```suggestion
          "description": "Describes files and/or sub-folders if existent",
```

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -78,14 +79,23 @@ final class OxfordNanoporeMeasurement {
...
-            this.pooledSamplesMeasurement = folders["fast5fail"] ? folders["fast5fail"].getChildren().get(0) instanceof Fast5Folder : false
+            this.pooledSamplesMeasurement = containsAtLeastOneBarcodedFolder(folders["fast5fail"])
         }
     }
 
+    private static boolean containsAtLeastOneBarcodedFolder(DataFolder folder) {
+        if (!folder) {
+            return false
+        } else if (folder.getChildren()) {
+            return folder.getChildren().get(0) instanceof Fast5Folder
```
KochTobi, line 94: This should go over all children and return at the first hit.

KochTobi: Alternatively rename the method to `containsBarcodedFolderAsFirstChild`

PR #34, opened 2020-07-17
Refactor #33

KochTobi:
Fix #33 

In general, the names of methods and their functionality should
always fit together. This was not given in the current case and
is fixed with this patch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.9.4
 
-* Fix implementation of `containsAtLeastOneBarcodedFolder`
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Release Changelog for the data model library
 
+## 1.9.4
+
+* Fix implementation of `containsAtLeastOneBarcodedFolder`
```
sven1103, line 5:
```suggestion
* Fix implementation of `containsAtLeastOneBarcodedFolder`, such that the method checks all child elements to contain at least one barcoded folder in order to flag the measurement as pooled measurement.
```

sven1103 (changes requested):
The CL should be precises of what it fixes or reference an issue on Github

KochTobi:
Already committed your suggestion ;)

sven1103 (approved):

PR #35, opened 2020-07-31
Add accounting data objects

KochTobi:
### CostItem
 * Cost items describe services and their price. They are the building blocks of invoices and offers.
 * The unit price is always provided in euros.
### CostEstimate
 * During project planning a cost estimate is often required to list expected costs.
 * This estimate is not intended to be legally binding. 
 * It provides an overview of the project scope and the costs associated with it.
### Offer
 * An offer describes a legally binding service proposal with associated costs.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -1,0 +1,23 @@
+package life.qbic.datamodel.accounting
+
+import life.qbic.datamodel.persons.Person
+import life.qbic.datamodel.projects.ProjectInfo
+
+/**
+ * A cost estimate for a project
+ *
+ * During project planning a cost estimate is often required to list expected costs.
+ * This estimate is not intended to be legally binding. It provides an overview over the project
+ * scope and the costs associated with it.
+ *
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostEstimate {
+    Date modificationDate
+    Person customer
+    ProjectInfo projectInfo
+    List<CostItem> items
+    double totalPrice
+}
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -20,4 +20,5 @@ class CostEstimate {
     ProjectInfo projectInfo
     List<CostItem> items
     double totalPrice
+    final String IDENTIFIER
 }
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
@@ -12,7 +12,6 @@ package life.qbic.datamodel.accounting
 class CostItem {
 
     double quantity
-    final String IDENTIFIER
     final double UNIT_PRICE
     final String UNIT_NAME
     final String NAME
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
@@ -1,0 +1,44 @@
+package life.qbic.datamodel.accounting
+
+/**
+ * An item with associated costs and quantity
+ *
+ * Cost items describe services and their price. They are the building blocks of invoices and offers.
+ * The unit price is always provided in euros.
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostItem {
+
+    double quantity
+    final String IDENTIFIER
+    final double UNIT_PRICE
+    final String UNIT_NAME
+    final String NAME
+    final String DESCRIPTION
+
+    CostItem(double UNIT_PRICE, String UNIT_NAME, String NAME, String DESCRIPTION) {
+        this.UNIT_PRICE = UNIT_PRICE
+        this.UNIT_NAME = UNIT_NAME
+        this.NAME = NAME
+        this.DESCRIPTION = DESCRIPTION
+    }
+
+    CostItem(double quantity, double UNIT_PRICE, String UNIT_NAME, String NAME, String DESCRIPTION) {
+        this.quantity = quantity
+        this.UNIT_PRICE = UNIT_PRICE
+        this.UNIT_NAME = UNIT_NAME
+        this.NAME = NAME
+        this.DESCRIPTION = DESCRIPTION
+    }
+
+    double getQuantity() {
+        return quantity
+    }
+
+    void setQuantity(double quantity) {
+        this.quantity = quantity
+    }
+}
+
--- a/src/main/groovy/life/qbic/datamodel/accounting/Offer.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/Offer.groovy
@@ -1,0 +1,30 @@
+package life.qbic.datamodel.accounting
+
+import life.qbic.datamodel.persons.Person
+import life.qbic.datamodel.projects.ProjectInfo
+
+/**
+ * An offer for a project
+ *
+ * An offer describes a legally binding service proposal with associated costs.
+ *
+ * @since: 1.9.0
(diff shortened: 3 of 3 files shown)


src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * An item with associated costs and quantity
+ *
+ * Cost items describe services and their price. They are the building blocks of invoices and offers.
+ * The unit price is always provided in euros.
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostItem {
```
jenniferboedker, line 12: NIT: maybe change to PackageItem to keep the vocabulary of the process?

KochTobi: Package item implies that the item is inside a package but it is actually the other way around as far as I understood it. For now, I only created a CostItem to keep it as simple as possible. We could differentiate further in the future if that is desired.

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * scope and the costs associated with it.
+ *
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostEstimate {
+    Date modificationDate
+    Person customer
+    ProjectInfo projectInfo
```
jenniferboedker, line 20: As far as I understand ProjectInfo in our context is project title and description. I would suggest to separate this from the openbis ProjectInfo as used here.

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * This estimate is not intended to be legally binding. It provides an overview over the project
+ * scope and the costs associated with it.
+ *
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostEstimate {
+    Date modificationDate
+    Person customer
```
jenniferboedker, line 19: fits very well :)

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+    double quantity
+    final double UNIT_PRICE
+    final String UNIT_NAME
```
jenniferboedker, line 16: nit: PACKAGE_TYPE

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
```diff
@@ -0,0 +1,43 @@
...
+ * Cost items describe services and their price. They are the building blocks of invoices and offers.
+ * The unit price is always provided in euros.
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class CostItem {
+
+    double quantity
+    final double UNIT_PRICE
```
jenniferboedker, line 15: NIT: PACKAGE_PRICE

jenniferboedker: or is it the price for the item?

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
jenniferboedker: or is it the type of the item?

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
```diff
@@ -0,0 +1,43 @@
...
+    final String UNIT_NAME
+    final String NAME
+    final String DESCRIPTION
+
```
jenniferboedker, line 19: maybe we need a list of packages of which this item consits?

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
jenniferboedker: As far as I understand it the price is calculated from the packages, therefore I would not set it with the constructor but have a method to calculate it? But then we have no POJO, right?

src/main/groovy/life/qbic/datamodel/accounting/Offer.groovy
```diff
@@ -0,0 +1,30 @@
...
+ * @since: 1.9.0
+ * @author: Tobias Koch
+ *
+ * * */
+class Offer {
+    Date modificationDate
+    Date expirationDate
+    Person customer
+    Person projectManager
+    ProjectInfo projectInfo
```
jenniferboedker, line 20: same as above :)

jenniferboedker (approved):
Nice, now we can start coding!:)

NIT:
However, as we already discussed: initially it was hard for me to understand the naming especially of the `CostItem` since we never used this term before. I would suggest to clearly communicate what is meant by it and maybe try to adjust the documentation to the new name (at least the architecture document) or adjust the naming to the designed architecture.

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
KochTobi: the unit price is the price per unit. so if the quantity is 2 then the total price for this item would be 2*unit_price

KochTobi: unit = package? probably. But as I mentioned the term package is very QBiC specific and not well defined.

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
KochTobi: This is `kg`, `MB`, `number`,...

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
KochTobi: I would try to keep it as simple as possible in the beginning. Therefore I tried to avoid the nesting of "packages" and "items".

src/main/groovy/life/qbic/datamodel/accounting/Offer.groovy
KochTobi: https://github.com/qbicsoftware/data-model-lib/pull/35/files#r463552367

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
jenniferboedker: ahhh this is the entity then... :D

KochTobi: It is one line here http://www.h2officesolutions.com/h2os_custguide/image997.jpg

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
jenniferboedker: True, but imo it is important to have the connection between the item and the packages

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
KochTobi: The [ProjectInfo](https://github.com/qbicsoftware/data-model-lib/blob/feature/qoffer2/src/main/java/life/qbic/datamodel/projects/ProjectInfo.java) already includes information about the project title, the secondary title, and description. Since this class represents information on project and only acts as a DTO (no logic), I would prefer to use the QBiC global definition.

src/main/groovy/life/qbic/datamodel/accounting/CostItem.groovy
KochTobi, line 19:
```suggestion
List<Overhead> overheads
```

Steffengreiner:
I'll close this PR since it's outdated. Holla @ me if you think this should still be included

PR #38, opened 2020-08-12
add electrone microscopy properties

jenniferboedker:
Adds the description and properties of electron microscopes


src/main/groovy/life/qbic/datamodel/dtos/imaging/parameters/electronmicroscopy/ElectronMicroscopyParameters.groovy
```diff
@@ -1,12 +1,35 @@
...
+    /**
+     * Shift of focus of the objective lens.
+     * It is used to observe 'Fresnel fringes', for taking lattice image or structure images.
+     */
+    private final Float targetDefocus
+
+    /**
+     * The accumulated amount of electrons transmitting the materials
+     */
+    private final Float cumulativeElectronDose
```
sven1103, line 34: Do we now sth about the common units?

jenniferboedker: I tried to be more specific, but i was unsure since the documentation on such properties is not so good / seems to vary between different microscopes

PR #52, opened 2020-08-27
Remove field access modifiers from DTOs

### LINKED ISSUE #51: DTO getter methods not generated
Groovy compilation does not generate getters when an access modifier is present on a field. This holds true for `private`, `protected`, and `public` class fields.

Declaring class fields without access modifiers generate getters and setters except for if the field is `final`. In this case, only getter methods are generated by groovy.

Since the fields of DTOs should be accessible, it would be nice to either remove the access modifier in groovy DTO classes or create getter methods explicitly.

KochTobi:
This addresses #51


Steffengreiner (approved):
LGTM! I find your solution pretty 💮 "groovy" 💮

PR #53, opened 2020-08-27
Hotfix 1.10.1

### LINKED ISSUE #51: DTO getter methods not generated

KochTobi:
Fixes #51 

EDIT: This PR is replaced by #54

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@
  
 # Maven
 target/
+.flattened-pom.xml
 
 # Some of our applications generate logs, which should be ignored by git
 logs/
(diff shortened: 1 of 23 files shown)


Steffengreiner (approved):
I'll go ahead and approve this PR as well! LGTM

PR #59, opened 2020-09-16
Account for unknown affiliation category

sven1103:


jenniferboedker (approved):
LGTM, though I think we have different values stored in our DB. E.g. Unkown is not included. I guess they should have the same values, right?

sven1103:
@jenniferboedker i think this is fine, we dont need to account this on the DB level. However in the source code, this would avoid the need of passing null, if the category value would not match.

PR #61, opened 2020-09-22
Fix customer tests

KochTobi:
This PR adjusts the existing tests to correct the test method names and use AcademicTitle enums for customer creation.


Steffengreiner (approved):
LGTM!

PR #64, opened 2020-09-29
Template merge

sven1103:
This PR introduces the initial project skeleton based on QUBE, which holds the current templates for QBiC software projects.

Future PRs will be smaller!

This allows us to enable syncing, once a new template version was released with QUBE. Thus, we can track changes in the template and incorporate them into QBiC software products, using the Github workflow `sync_project`, that is provided with this PR (`.github/sync_project.yml`).


Zethson:
@sven1103 since the sync workflow is failing with  a credentials error - how did you setup the secret?

Zethson:
+ Note that the TEMPLATE branch should have never seen the sync workflow.

We may run into self modifying workflow issues elsewise.

Zethson:
@sven1103 Yeah I thought about using Jonny for sync tasks. Since QBiC is also "centralized" for any qbicsoftware repositories this should work.

-> If we use Jonny no QBiC member can create repositories outside of qbicsoftware with sync support (e.g. PhD students)
-> If this is fine we can hardcore jonny + his secret.

Thoughts?

.github/workflows/sync_project.yml
```diff
@@ -0,0 +1,30 @@
...
+          with:
+              fetch-depth: 0
+          name: Check out source-code repository
+
+        - uses: oleksiyrudenko/gha-git-credentials@v1
+          with:
+               name: 'sven1103'
+               email: 'sven.fillinger@qbic.uni-tuebingen.de'
+               actor: 'sven1103'
+               token: '${{ secrets.QUBE_SYNC_TOKEN}}'
```
jenniferboedker: Is this the standard then? Or does this need to be updated by everyone pulling the code?

Zethson: @jenniferboedker We will hardcode Johnny and you do no need to adapt anything.

jenniferboedker (approved):
LGTM, but solely checked who concise files describing how PRs are written. 
Since my experience in Github Actions is low I would leave that to @Zethson.

.github/workflows/sync_project.yml
jenniferboedker: Ah ok, so Johnny is the developer of QBiC in Github?

Zethson: Johnny is the best and most hardworking developer of QBiC. 
He never complains and just does what he is supposed to do (most of the times).

PR #65, opened 2020-09-29
Add equals and hashValue methods for each dto class

KochTobi:
The DTOs did not provide `equals` and `hashValue` methods. This PR adds those methods with the assumption that all fields can be null since the classes in question are DTOs and only provide the data fields.


src/main/groovy/life/qbic/datamodel/dtos/business/Affiliation.groovy
```diff
@@ -42,7 +42,35 @@ final class Affiliation {
...
+    if (organisation != that.organisation) return false
+    if (postalCode != that.postalCode) return false
+    if (street != that.street) return false
+
+    return true
+  }
+
+  int hashCode() {
+    int result
+    result = (organisation != null ? organisation.hashCode() : 0)
```
jenniferboedker, line 64: why not 
```suggestion
    result = 31* result + (organisation != null ? organisation.hashCode() : 0)
```

src/main/groovy/life/qbic/datamodel/dtos/imaging/Image.groovy
```diff
@@ -101,4 +101,37 @@ final class Image {
...
+    if (modality != image.modality) return false
+    if (pixelSize != image.pixelSize) return false
+    if (resolution != image.resolution) return false
+
+    return true
+  }
+
+  int hashCode() {
+    int result
+    result = (acquisitionDate != null ? acquisitionDate.hashCode() : 0)
```
sven1103, line 126: Elvis ftw
```suggestion
    result = ( acquisitionDate?.hashCode() ?: 0)
```

sven1103 (changes requested):
Please use the elvis operator and null safe operator, that is syntactic sugar of Groovy to avoid null checks.

And here, its application makes a lot of sense and improves readability of the code.

src/main/groovy/life/qbic/datamodel/dtos/imaging/Image.groovy
KochTobi: I will do that in a separate commit. The solution above is the default intellij groovy hashCode method.

KochTobi:
> Please use the elvis operator and null safe operator, that is syntactic sugar of Groovy to avoid null checks.
> 
> And here, its application makes a lot of sense and improves readability of the code.

Do you think we could also use 
```
 import groovy.transform.EqualsAndHashCode
 @EqualsAndHashCode
```

sven1103 (approved):
400 less lines of code, i like :)

PR #66, opened 2020-10-01
Implement AffiliationCategoryFactory

KochTobi:
This PR implements a factory to get an `AffiliationCategory` from a string value.


sven1103 (approved):
LGTM

PR #67, opened 2020-10-02
Add identifiers for TOMATO

jenniferboedker:
**Purpose of PR**
Offers generated with TOMATE must be distinguishable, therefore this PR introduces identifiers for this portlet.

**Change**
To describe the identifier the classes `TomatoId`, `QuotationId` and `CostEstimateId` have been introduced. 
TOMATO offers two types of offers which are reflected with `TomatoId` as an abstract class. It defines how the deriving quotation and cost estimate needs to look like. It also handles the logic for creating the identifier.

**Desired Feedback**
Feel free to critique my approach

- [-] Quick pair of eyes
- [x] Discussion on technical approach
- [x] Critique on design
- [-] Review of copy


src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A DTO describing Tomato identifiers
+ *
+ * The TomatoId should be used as a general class describing identifiers like the {@link QuotationId} and the {@link CostEstimateId}.
+ * It consists of a random part, a project conserved part and a version. Furthermore, the type is also part of the identifier.
+ *
+ * @since: 1.11.0
+ *
+ */
+abstract class TomatoId {
```
jenniferboedker, line 12: Or name it OfferManagerId?

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+    /**
+     * Assembles the identifier from the class attributes. This method determines how the identifier
+     * looks like for all implementing classes to provide a uniform identifier format
+     *
+     * @return a String containing the type, project conserved part, random part and the version
+     */
+    String getIdentifier(){
+        return getType() + "_" + getProjectConservedPart() + "_" + getRandomPart() + "_" + "v" + getVersion()
+    }
```
jenniferboedker, line 53: The identifier is created here. Do we want to have it in the DTO since it is logic that should not be part of it?

jenniferboedker:
[Provide a public API for interaction with the Offer/CE ID DTOs](https://trello.com/c/OVwS83xB/240-provide-a-public-api-for-interaction-with-the-offer-ce-id-dtos)

jenniferboedker:
[Create an CE & Offer ID DTOs in the data-model-lib](https://trello.com/c/tt9gtnhk/239-create-an-ce-offer-id-dtos-in-the-data-model-lib)

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    String getRandomPart() {
+        return randomPart
+    }
+
+    @Override
+    int getVersion() {
+        return version
+    }
+}
```
KochTobi, line 59: I do not think you need those.
```suggestion
```

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
```diff
@@ -0,0 +1,57 @@
...
+     */
+    private String projectConserved
+    /**
+     * Random part of the identifier
+     */
+    private String random
+    /**
+     * Version of the identifier
+     */
+    private int version
```
KochTobi, line 27: You could also use `{@InheritDoc}` here

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
```diff
@@ -0,0 +1,57 @@
...
+
+    @Override
+    String getRandomPart() {
+        return random
+    }
+
+    @Override
+    int getVersion() {
+        return version
+    }
```
KochTobi, line 56: I think you do not need those.
```suggestion
```

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,81 @@
...
+     * @return
+     */
+    abstract String getRandomPart()
+
+    /**
+     * Returns the version of the identifier
+     * @return
+     */
+    abstract int getVersion()
+}
```
KochTobi, line 81: You could provide implementations here since as a default behaviour. I would argue that you do even not need to implement getters since you can just leave the fields public (It is a DTO so this is the expected behaviour)

KochTobi (commented):
Some minor comments. Overall looks good

KochTobi (commented):
I just noticed that your `@since` is not up to date. It should be `1.12.0` I guess.

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
```diff
@@ -0,0 +1,59 @@
...
+    private int version
+
+
+    CostEstimateId(String projectConservedPart, String randomPart, int version) {
+        super(type,projectConservedPart,randomPart,version)
+
+        this.projectConserved = projectConservedPart
+        this.random = randomPart
+        this.version = version
+    }
```
KochTobi, line 37: `type` is not known in this scope
```suggestion
    CostEstimateId(String projectConservedPart, String randomPart, int version) {
        String type = "CE"
        super(type,projectConservedPart,randomPart,version)
    }    
    private CostEstimateId(String type, String projectConservedPart, String randomPart, int version) {
        super(type,projectConservedPart,randomPart,version)
        this.type = type
        this.projectConserved = projectConservedPart
        this.random = randomPart
        this.version = version
    }
```

KochTobi (changes requested):

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
KochTobi: What about AccountingId ?

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
KochTobi: I would argue against having it in the DTOs
There already was/is an `accounting` package

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
jenniferboedker: I would opt for keeping it as an attribute since otherwise, we could also create a QuotationId by just entering Q as type.
I see the type as a characteristic of the class. Why is type not known in this space?

jenniferboedker:
> I just noticed that your `@since` is not up to date. It should be `1.12.0` I guess.

Oh my bad, will fix it. Thanks for pointing it out.

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
KochTobi: I am not against keeping it. It simply does not compile this way

jenniferboedker: Fixed :) 
e71ae379f

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
jenniferboedker: But I thought it would be nicer to know what fields are expected for the abstract class. This is how you make variables in abstract classes to be implemented. Otherwise, I would have no control over the methods/variables implemented by other identifier classes.

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
jenniferboedker: See my comment about why the methods are required for abstract classes and therefore they need to be implemented by the implementing class :)

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
jenniferboedker: See my comment about why the methods are required for abstract classes and therefore they need to be implemented by the implementing class :)

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
jenniferboedker: OfferManagerId it is 45e98933d :)

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
jenniferboedker: Do you mean because of its logic is part of a DTO? But on the other hand, then I would argue to implement it in the offer manager itself.
How do the others feel about this?

jenniferboedker: @qbicsoftware/itss

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
KochTobi: At the moment you are shadowing the fields of the abstract class in your subclasses. This is unnecessary. You can implement it in the abstract class and use it or move it to the subclasses.

KochTobi: You only overwrite if you want to change behavior.

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
KochTobi: I responded in another comment

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
KochTobi: I resonded in another comment.

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -1,5 +1,7 @@
 package life.qbic.datamodel.accounting
 
+import life.qbic.datamodel.dtos.business.CostEstimateId
+import life.qbic.datamodel.dtos.business.QuotationId
```
KochTobi, line 4: Just to note that down here:
at this point, the objects become business objects and are not simply DTOs anymore.

src/main/groovy/life/qbic/datamodel/accounting/Quotation.groovy
```diff
@@ -11,7 +13,7 @@ import life.qbic.datamodel.persons.Person
  * @author: Tobias Koch
  *
  */
-class Offer {
+class Quotation {
```
KochTobi, line 16: Nit: I remember now that @sven1103 and I discussed this and he mentioned that an offer and a quotation are two separate things in terms of when they are created, paid, and so on.

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimateId.groovy
```diff
@@ -0,0 +1,29 @@
...
+
+    CostEstimateId(String projectConservedPart, String randomPart, int version) {
+        super(type,projectConservedPart,randomPart,version)
+    }
+
+
+    @Override
+    String getType() {
+        return type
+    }
```
KochTobi, line 27: You can also drop this method. The getter is the same for every subclass and the behavior is no different than for the superclass. Also, you do implement a getter to a static variable here...

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
```diff
@@ -0,0 +1,27 @@
...
+
+    QuotationId(String projectConservedPart, String randomPart, int version) {
+        super(type,projectConservedPart,randomPart,version)
+    }
+
+    @Override
+    String getType() {
+        return type
+    }
+
```
KochTobi, line 26:
```suggestion
```

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,87 @@
...
+    String getIdentifier(){
+        return getType() + "_" + getProjectConservedPart() + "_" + getRandomPart() + "_" + "v" + getVersion()
+    }
+
+    /**
+     * Returns the type of the identifier
+     *
+     * @return
+     */
+    abstract String getType()
```
KochTobi, line 60:
```suggestion
   String getType() {
       return type
   }
```

KochTobi (changes requested):
Overall looks very good. The only thing I would ask is to also move the type getter to the super class.

src/main/groovy/life/qbic/datamodel/accounting/Quotation.groovy
jenniferboedker: Correct, but this will reflect the Quotation right, I don't see why this should be an Offer but have a QuotationId. There we would have confusion in naming again. At least imo.

src/main/groovy/life/qbic/datamodel/dtos/business/QuotationId.groovy
jenniferboedker: If I remove the getter Id cannot retrieve the type from the extending classes

jenniferboedker: How should my abstract class know about the type variable in the extending classes?

KochTobi: The problem here is that the static variable shadows the class field of the abstract class.
Suggestion: remove the static variable and just call the constructor with "Q"

KochTobi: > How should my abstract class know about the type variable in the extending classes?

I fear I do not understand your question. How is this field different from all the other fields?

src/main/groovy/life/qbic/datamodel/accounting/Quotation.groovy
KochTobi: Right the question then is in whether we should have an OfferId instead of a QuotationId 🤷

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    /**
+     * Returns the version of the identifier
+     * @return
+     */
+    int getVersion() {
+        return version
+    }
+}
```
KochTobi, line 80:
```suggestion
    /**
     * Returns the type of the identifier
     * @return
     */
    int getType() {
        return type
    }
}
```
Since you deleted the getter from the subclasses you should implement it here. Otherwise, this field is obsolete.

src/main/groovy/life/qbic/datamodel/dtos/business/TomatoId.groovy
```diff
@@ -0,0 +1,80 @@
...
+     * @param type describing the type of the underlying identifier
+     * @param projectConservedPart describes the project conserved part of the identifier
+     * @param randomPart describes the random part of the identifier
+     * @param version describes the version of the identifier
+     */
+    TomatoId(String type, String projectConservedPart, String randomPart, int version){
+        this.type = type
+        this.projectConserved = projectConservedPart
+        this.random = randomPart
+        this.version = version
```
KochTobi, line 42: Think about requiring the fields to be non-null.
```suggestion
        this.type = Objects.requireNonNull(type, "type must not be null")
        this.projectConserved = Objects.requireNonNull(projectConservedPart, "projectConservedPart must not be null")
        this.random = Objects.requireNonNull(randomPart, "randomPart must not be null")
        this.version = Objects.requireNonNull(version, "version must not be null")
```

KochTobi (changes requested):

KochTobi (approved):
Looks very good now @jenniferboedker :+1:

PR #70, opened 2020-10-02
Remove empty example package

KochTobi:
This PR removes an empty Example class that was contained in an example package. The package was initially committed due to a template branch merge.


sven1103 (approved):
thanks!

PR #80, opened 2020-11-02
Update offer related classes

jenniferboedker:
**Purpose of this PR**
Prepare the classes in the datamodel-lib to be used within TOMATO

**Changes**
With this PR the new class Person (in life/qbic/datamodel/dto/general) is added. It is a copy of the current Customer class (in life/qbic/datamodel/dto/business). This should allow delaying the decision on how to differentiate between customer and project manager. I also added the variable person type to enable the separation of different persons.

Furthermore, some functions in the Quotation and CE classes are implemented in order to be able to use them now in TOMATO.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -1,8 +1,7 @@
 package life.qbic.datamodel.accounting
 
 import life.qbic.datamodel.dtos.business.CostEstimateId
-import life.qbic.datamodel.dtos.business.QuotationId
-import life.qbic.datamodel.persons.Person
+import life.qbic.datamodel.dtos.general.Person
 
 /**
  * A cost estimate for a project
@@ -60,5 +59,27 @@ class CostEstimate {
         this.items = items
         this.totalPrice = totalPrice
         this.identifier = identifier
+    }
+
+    /**
+     * Adds a new item to the items list of the offer
+     *
+     * @param item which should be added to current list of items
+     */
+    public void addItem(ProductItem item){
+        items.add(item)
+    }
+
+    /**
+     * Removes an item from the list of items of the offer
+     *
+     * @param item which should be removed from the current list of items
+     */
+    void removeItem(ProductItem item){
+        items.each {
+            if(it.getNAME() == item.getNAME() && it.getProduct() == item.getProduct()){
+                items.remove(it)
+            }
+        }
     }
 }
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -76,10 +76,6 @@ class CostEstimate {
      * @param item which should be removed from the current list of items
      */
     void removeItem(ProductItem item){
-        items.each {
-            if(it.getNAME() == item.getNAME() && it.getProduct() == item.getProduct()){
-                items.remove(it)
-            }
-        }
+         items.remove(item)
     }
 }
--- a/src/main/groovy/life/qbic/datamodel/accounting/ProductCategory.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/ProductCategory.groovy
@@ -12,10 +12,10 @@ package life.qbic.datamodel.accounting
 enum ProductCategory {
 
     SEQUENCING("Sequencing"),
-    PROJECTMANAGEMENT("Project Management"),
-    PRIMARYBIOINFO("Primary Bioinformatics"),
-    SECONDARYBIOINFO("Secondary Bioinformatics"),
-    DATASTORAGE("Data Storage")
+    PROJECT_MANAGEMENT("Project Management"),
+    PRIMARY_BIOINFO("Primary Bioinformatics"),
+    SECONDARY_BIOINFO("Secondary Bioinformatics"),
+    DATA_STORAGE("Data Storage")
 
     /**
      * Value describing the enum type with a string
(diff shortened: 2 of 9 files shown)


src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
```diff
@@ -45,11 +45,11 @@ class ProductItem {
     }
 
     /**
-     * Sums up the cost of the item based on its quantity and the price and unit defined in the product
+     * Calculates the cost of the item based on its quantity and the price and unit defined in the product
      * @return
      */
     double computeTotalCosts(){
-        //todo implement me
+        return product.unitPrice * quantity
```
KochTobi, line 52: Are overheads also included here? Or is this not the right place to put this?

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -61,4 +60,26 @@ class CostEstimate {
...
+     * Removes an item from the list of items of the offer
+     *
+     * @param item which should be removed from the current list of items
+     */
+    void removeItem(ProductItem item){
+        items.each {
+            if(it.getNAME() == item.getNAME() && it.getProduct() == item.getProduct()){
+                items.remove(it)
+            }
+        }
```
KochTobi, line 83:
```suggestion
         items.remove(item)
```
The `equals` method should be implemented in the `ProductItem`

src/main/groovy/life/qbic/datamodel/accounting/Quotation.groovy
```diff
@@ -87,7 +86,11 @@ class Quotation {
      * @param item which should be removed from the current list of items
      */
     void removeItem(ProductItem item){
-        //todo implement me
+        items.each {
+            if(it.getNAME() == item.getNAME() && it.getProduct() == item.getProduct()){
+                items.remove(it)
+            }
+        }
```
KochTobi, line 93:
```suggestion
        items.remove(item)
```
The `equals` method should be implemented in the `ProductItem`

src/main/groovy/life/qbic/datamodel/dtos/business/Customer.groovy
```diff
@@ -12,62 +13,16 @@ import groovy.transform.EqualsAndHashCode
...
-  /**
-   * The customer's email address
-   */
-  final String eMailAddress
-
-  /**
-   * Associated affiliations
-   */
-  final List<Affiliation> affiliations
+  private String type = "customer"
```
KochTobi, line 18:
```suggestion
  private static final String TYPE = "customer"
```

KochTobi (changes requested):
Looks really good. However I would suggest to implement the `equals` method in the `ProductItem` class rather than implementing it in every class using the `ProductItem`. This can be done using the `EqualsAndHashCode` annotation. 
https://docs.groovy-lang.org/latest/html/api/groovy/transform/EqualsAndHashCode.html

src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
jenniferboedker: I was not sure about it. I just put the implementation that was obvious for me. We can still change it when we realize we need another calculation here, right?

KochTobi: Sure. We just need to keep this in mind. It is also the most intuitive way to calculate the price of a product item.

jenniferboedker: Right, we also need to think about where we want to place more specific calculations. Could also imagine that overheads and stuff could be added directly in the use case

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -61,4 +60,22 @@ class CostEstimate {
...
+    }
+
+    /**
+     * Removes an item from the list of items of the offer
+     *
+     * @param item which should be removed from the current list of items
+     */
+    void removeItem(ProductItem item){
+         items.remove(item)
+    }
```
KochTobi, line 80: Nice JavaDoc! I would argue that those methods belong to the super class `TomatoId` and are to be overwritten here only in case the behaviour differs to the super class. Quotation has the same methods.

KochTobi (changes requested):
Nice work overall. 
I would ask you to move the `addItem` and `removeItem` methods to the superclass.

Other than that I expect the current version to be not usable because of an assignment issue in the `TomatoId` class where the `int getType()` is expected to return an `int` from a `String type` field. Since this is not part of this PR but already is an error in the code base I could also open an issue if you would prefer that.

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
jenniferboedker: For the quotation and the cost estimate exists no superclass. Just for their IDs. And these functions are not part of the ID classes .. ;)

KochTobi: Oh I see. Should there be a super class?

src/main/groovy/life/qbic/datamodel/accounting/ProductCategory.groovy
```diff
@@ -12,10 +12,10 @@ package life.qbic.datamodel.accounting
...
 
     SEQUENCING("Sequencing"),
-    PROJECTMANAGEMENT("Project Management"),
-    PRIMARYBIOINFO("Primary Bioinformatics"),
-    SECONDARYBIOINFO("Secondary Bioinformatics"),
-    DATASTORAGE("Data Storage")
+    PROJECT_MANAGEMENT("Project Management"),
+    PRIMARY_BIOINFO("Primary Bioinformatics"),
+    SECONDARY_BIOINFO("Secondary Bioinformatics"),
+    DATA_STORAGE("Data Storage")
```
KochTobi, line 18: A thousand :+1:

KochTobi (approved):
LVGTM = Looks very good to me

PR #81, opened 2020-11-03
Rename Quotation to Offer and QuotationId to OfferId. Additionally Ch…

Steffengreiner:
…ange String Type of OfferId from "Q" to "O"


jenniferboedker (approved):
lgtm

PR #82, opened 2020-11-04
Clean Up commented code from Address DTO and replace null values with

### LINKED ISSUE #55: life.qbic.datamodel.people.Address can contain null
The [Address class](https://github.com/qbicsoftware/data-model-lib/blob/master/src/main/groovy/life/qbic/datamodel/people/Address.groovy) has a default constructor and can return null on values leading to NullPointerExceptions. Example for an address containing null below.

```
long imageId = imageInfo.getImageId();
ImagingHardware hardware =new SRMHardware.Builder("some objective", new Detector("dummy detector")).build();
Location dummyLocation = new Location("13", new Address());
Instrument dummyInstrument = new Instrument("test_manufacturer", "Donnerblitz 2001", "db-2001-hp.5", dummyLocation, "DUMMY", hardware);
omeroClient.associateInstrumentWithImage(dummyInstrument, imageId);
```

It would be better to set default values of empty Strings using the default constructor or a builder pattern to avoid Nullpointers from occurring in the DTO.

Steffengreiner:
default values for each property. 
This is intended as a fix for the Issue reported by @KochTobi. 
See #55

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/people/Address.groovy
+++ b/src/main/groovy/life/qbic/datamodel/people/Address.groovy
@@ -4,25 +4,25 @@ import com.fasterxml.jackson.annotation.JsonProperty
 
 class Address  {
 
-  private String affiliation
-  private String street
-  private Integer zipCode
-  private String country
+  private String affiliation = ""
+  private String street = ""
+  private Integer zipCode = 0
+  private String country = ""
 
   /**
    * Get affiliation
    * @return affiliation
    **/
   @JsonProperty("affiliation")
-  public String getAffiliation() {
+  String getAffiliation() {
     return affiliation;
   }
 
-  public void setAffiliation(String affiliation) {
+  void setAffiliation(String affiliation) {
     this.affiliation = affiliation;
   }
 
-  public Address affiliation(String affiliation) {
+  Address affiliation(String affiliation) {
     this.affiliation = affiliation;
     return this;
   }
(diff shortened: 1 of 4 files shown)


src/main/groovy/life/qbic/datamodel/people/Address.groovy
```diff
@@ -86,33 +68,31 @@ class Address  {
...
+  String toString() {
     StringBuilder sb = new StringBuilder();
     sb.append("class Address {\n");
-
     sb.append("    affiliation: ").append(toIndentedString(affiliation)).append("\n");
     sb.append("    street: ").append(toIndentedString(street)).append("\n");
-//    sb.append("    number: ").append(toIndentedString(number)).append("\n");
     sb.append("    zipCode: ").append(toIndentedString(zipCode)).append("\n");
     sb.append("    country: ").append(toIndentedString(country)).append("\n");
     sb.append("}");
```
KochTobi, line 98: I think to keep it consistent we could also move to the Annotation and delete the equals method
https://docs.groovy-lang.org/latest/html/api/groovy/transform/EqualsAndHashCode.html

src/main/groovy/life/qbic/datamodel/people/Address.groovy
```diff
@@ -4,25 +4,25 @@ import com.fasterxml.jackson.annotation.JsonProperty
...
     return affiliation;
   }
 
-  public void setAffiliation(String affiliation) {
+  void setAffiliation(String affiliation) {
     this.affiliation = affiliation;
   }
 
-  public Address affiliation(String affiliation) {
+  Address affiliation(String affiliation) {
```
KochTobi, line 25: This looks wrong.

KochTobi (changes requested):
Overall really good idea. My question would be if an address should be modifiable or if we should move to a final class with Builder pattern here. If the former would be the case I'd suggest removing all getter and setter methods and letting groovy generate them.

What do you think?

Steffengreiner:
As previously discussed, i introduced a seperate "Address" DTO with a builder pattern. 
The old address is set as deprecated but got some default values to avoid throwing NullExceptions.

sven1103 (approved):
LGTM

src/main/groovy/life/qbic/datamodel/people/Address.groovy
KochTobi:
```suggestion
  Affiliation affiliation(String affiliation) {
```
Still relevant.

sven1103: No, it is correct. It is how fluent APIs have been designed for mutable objects. 

And the return type must not change, as it would break the API contract.

KochTobi: Ah so it is like a Builder pattern without a builder?

src/main/groovy/life/qbic/datamodel/people/Address.groovy
KochTobi: Would also save us from 1 or 2 letter variable names ;)

src/main/groovy/life/qbic/datamodel/people/Address.groovy
sven1103: old fashioned way, yes.

src/main/groovy/life/qbic/datamodel/people/Address.groovy
Steffengreiner: Really Good Feedback! Based on it, I introduced a seperate "Address" DTO with a builder pattern. 
After internal discussions, we agreed that it's alright to set the old implementation to deprecated to avoid breaking dependencies down the line and remove it in a future PR.

KochTobi (approved):

PR #84, opened 2020-11-06
Add affiliation attribute to offer

jenniferboedker:
**Purpose of this PR**
Currently, it is not possible to calculate the offer price with the information provided in the Offer DTO since we do not have a field where we decide on a fixed affiliation for the customer. We only have a list of affiliations and thus cannot decide on the affiliation category to determine the overheads for the customer.

**Changes**
 I would suggest adding a field `selectedCustomerAffiliation` in the offer DTO since the affiliation of the customer is only fixed in the context of an offer.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
@@ -49,6 +49,10 @@ class Offer {
      * The identifier for the offer which makes it distinguishable from other offers
      */
     final OfferId identifier
+    /**
+     * The affiliation of the customer selected for this offer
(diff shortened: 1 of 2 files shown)


KochTobi (approved):
LGTM

However, we should think about the Builder Pattern in the Offer class since it does not add functionality when we always need all information to be present to create an Offer.

jenniferboedker:
I agree, should we add an issue for this or how do you think we can keep track it?

KochTobi:
> I agree, should we add an issue for this or how do you think we can keep track it?

Yes do you want to create one or should I?

jenniferboedker:
done :) #85

PR #86, opened 2020-11-06
Feature/project manager

Steffengreiner:
**Purpose of this PR**
This PR introduces the `ProjectManager` dto as an extension of the abstract `Person` class. 
It's intended to store the information of a `project manager` whose e.g. associated with an Offer. 
Additionally it removes the redundant `persontype` property from the `Person `class and it's implementation in the `customer` dto. 
**Misc**
I'm unsure, if we should set a default value for the projectManagers `affiliations` and `AcademicTitle` property or how they should be handled.


KochTobi (commented):
LGTM

Only 1 question: Since we have a DTO here shouldn't it be fine for the values to be null?

Steffengreiner:
@KochTobi Good question, the `requireNonNull` property for the properties of the abstract `Person `class was already set before. I'd have no issue removing it since as of now it's only used in the TOMATO Context where we make sure that null values are handled properly.

src/main/groovy/life/qbic/datamodel/dtos/general/Person.groovy
```diff
@@ -46,13 +41,11 @@ abstract class Person {
    */
   final List<Affiliation> affiliations
 
-  Person(String personType,
-         String firstName,
+  Person(String firstName,
```
sven1103, line 44: This change breaks all implementations of the Customer class. 

It would be preferable to mark it as @deprecated and introduce a Builder without the type instead

src/test/groovy/life/qbic/datamodel/dtos/business/ProjectManagerSpec.groovy
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        def "missing affiliation list shall cause the projectManager to contain an empty list"() {
+            when:
+            def projectManager = new ProjectManager("", "", AcademicTitle.NONE, "a.b@c.de", null)
+            then:
+            projectManager.affiliations != null
+            projectManager.affiliations.isEmpty()
+        }
+    }
```
KochTobi, line 43: I think the tests need to be adjusted or removed since this is no longer expected behavior.

KochTobi (dismissed):
LGTM I would remove the tests that do not test for expected behaviour (NullPointerException)

sven1103 (approved):

PR #87, opened 2020-11-06
Refactor/cost estimate and product item

Steffengreiner:
**Purpose of this PR**
This PR refactors the `CostEstimate` Class to the same builder pattern as the `Offer` Class. 
Additionally it adds the previously missing fields `ExpirationDate` and `ProjectManager` to the `CostEstimate `class 
and removes the ` computeTotalCost `method  from the `ProductItem` class making `ProductItem` a proper dto.
Finally it introduces simple tests for the introduced and changed classes. 

**Misc**
This PR also deprecates the `Person` class to mark it for future removal, since i was unsure if that tag in itself would be worth a seperate PR.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -1,7 +1,8 @@
 package life.qbic.datamodel.accounting
 
+import life.qbic.datamodel.dtos.business.Affiliation
 import life.qbic.datamodel.dtos.business.CostEstimateId
-import life.qbic.datamodel.dtos.general.Person
+import life.qbic.datamodel.dtos.business.Customer
 
 /**
  * A cost estimate for a project
@@ -22,9 +23,19 @@ class CostEstimate {
     final Date creationDate
 
     /**
+     * The date on which the cost estimate expires
+     */
+    final Date expirationDate
+
+    /**
      * The customer for which the estimate is created
      */
-    final Person customer
+    final Customer customer
+
+    /**
+     * The QBiC project manager who was assigned to the project
+     */
+    final ProjectManager projectManager
 
     /**
      * The title of the project
@@ -51,14 +62,21 @@ class CostEstimate {
      */
     final CostEstimateId identifier
 
-    CostEstimate(Date creationDate, Person customer, String projectTitle, String projectDescription, List<ProductItem> items, double totalPrice, CostEstimateId identifier) {
+    /**
+     * An identifier which makes the CostEstimate distinguishable from other CostEstimates
+     */
+    final Affiliation selectedCustomerAffiliation
+
+    CostEstimate(Date creationDate, Customer customer, ProjectManager projectManager, String projectTitle, String projectDescription, List<ProductItem> items, double totalPrice, CostEstimateId identifier, Affiliation selectedCustomerAffiliation) {
         this.creationDate = creationDate
         this.customer = customer
+        this.projectManager = projectManager
         this.projectTitle = projectTitle
         this.projectDescription = projectDescription
         this.items = items
         this.totalPrice = totalPrice
         this.identifier = identifier
+        this.selectedCustomerAffiliation = selectedCustomerAffiliation
     }
 
     /**
--- a/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
+++ b/src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
@@ -3,7 +3,7 @@ package life.qbic.datamodel.accounting
 import life.qbic.datamodel.dtos.business.Affiliation
 import life.qbic.datamodel.dtos.business.CostEstimateId
 import life.qbic.datamodel.dtos.business.Customer
-import life.qbic.datamodel.dtos.general.Person
+import life.qbic.datamodel.dtos.business.ProjectManager
 
 /**
  * A cost estimate for a project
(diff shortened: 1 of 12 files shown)


src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -19,12 +21,22 @@ class CostEstimate {
...
     /**
      * The customer for which the estimate is created
      */
-    final Person customer
+    final Customer customer
+
+    /**
+     * The QBiC project manager who was assigned to the project
+     */
+    final Person projectManager
```
jenniferboedker, line 39: In which order do you plan to merge the PRs of yours. The one introducing the projectmanager object should be either way reflected here at some point. In this PR or in the other one so that we won't miss it :)

src/main/groovy/life/qbic/datamodel/people/Person.groovy
```diff
@@ -1,5 +1,6 @@
 package life.qbic.datamodel.people;
 
+@Deprecated
```
jenniferboedker, line 3: Did we want to deprecate this Person class? I think we don't even use it in the context of TOMATO.

src/test/groovy/life/qbic/datamodel/dtos/business/ProductItemSpec.groovy
```diff
@@ -0,0 +1,39 @@
...
+
+        then:
+        productItem.NAME == "Pandoras Box"
+        productItem.DESCRIPTION == "Don't open it!"
+        productItem.product == product
+    }
+
+    def "ProductItem shall store and provide the given properties: quantity, name, description and product"() {
+        when:
+        def productItem = new ProductItem(1, "Pandoras Box", "Don't open it!", product)
```
jenniferboedker, line 29: What happens when I open the box? 👀

jenniferboedker (changes requested):
I added some comments especially because of the choice of the Person class to be deprecated, is it the correct one? Furthermore, I think we should try to structure all of your PRS since we need to include the ProjectManager class also into the new CostEstimate.

After we decided on how to approach on how to merge your PRs I will approve.

Other than that, nice work! LGTM

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
Steffengreiner: Good Question, I was unsure how to proceed here since the projectManager Class is in a seperate PR and i wanted to make sure this PR works properly in it's current state. 
I'd rather merge the other pr into development and merge that back into this PR and adapt the property to make sure everything runs properly.

src/main/groovy/life/qbic/datamodel/people/Person.groovy
Steffengreiner: I think it would be a good idea to deprecate classes with the same name but different implemenations to avoid confusion in the future. I don't know in which other classes this class is used, therefore we discussed deprecating it and removing it as soon as we know how easy it is to do so

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
jenniferboedker: Thats a good idea, then I will wait for your merge until my final approval :)

src/test/groovy/life/qbic/datamodel/dtos/business/ProductItemSpec.groovy
Steffengreiner: Then all the IDEs in the world will stop working and we'll have to code in the terminal directly 😨

src/main/groovy/life/qbic/datamodel/accounting/CostEstimate.groovy
```diff
@@ -19,12 +21,22 @@ class CostEstimate {
...
     /**
      * The customer for which the estimate is created
      */
-    final Person customer
+    final Customer customer
+
+    /**
+     * The QBiC project manager who was assigned to the project
+     */
+    final ProjectManager projectManager
```
jenniferboedker, line 39: 👍

src/test/groovy/life/qbic/datamodel/dtos/business/CustomerSpec.groovy
```diff
@@ -36,10 +36,10 @@ class CustomerSpec extends Specification{
 
   def "missing affiliation list shall cause the customer to contain an empty list"() {
     when:
-    def customer = new Customer("", "", AcademicTitle.NONE, "a.b@c.de", null)
+    def projectManager = new ProjectManager("", "", AcademicTitle.NONE, "a.b@c.de", null)
     then:
-    customer.affiliations != null
-    customer.affiliations.isEmpty()
+    projectManager.affiliations != null
+    projectManager.affiliations.isEmpty()
```
jenniferboedker, line 42: Since this is the customer spec I would keep the customer here.. if you dont want to keep the customer here I would suggest to update the comment :)

jenniferboedker: I would keep the old test, as I see you created a ProjectManagerSpec so no need to include the project manager to the customer spec

jenniferboedker (commented):
Left some minor comment about your change for one test, if you address this, this PR is approved 👍  good work

src/test/groovy/life/qbic/datamodel/dtos/business/CustomerSpec.groovy
Steffengreiner: No one saw that... >< Nice Catch!

jenniferboedker (approved):
LGTM

PR #89, opened 2020-11-10
Refactor/persons

sven1103:
Introduces an abstract builder pattern and two example implementations for a common person and a project manager.


Steffengreiner (commented):
Very nice Job, the Builder Pattern looks clean! Travis however seems to doesn't like it since it fails with the error message: `type Builder does not take parameters"
I'll check it out and approve the PR when this is fixxed 👍

Steffengreiner (approved):
LGTM!

PR #95, opened 2020-11-13
Add product unit enum factory

sven1103:
This PR adds a product unit factory, that delivers a matching product unit enum for a given String value.


jenniferboedker (approved):
LGTM

Issue #85, opened 2020-11-06
BuilderPattern for Offer DTO's may be not suitable


jenniferboedker:
**Current Situation**
Offers are created through the builder pattern

**Purpose for change**
We always need all information to be present to create an Offer, this might be impractical.

PR #97, opened 2020-11-13
Refactor Offer DTO fields into required and optional parameters

Steffengreiner:
**Purpose of this PR**
This PR refactors the fields of the Offer DTO into required and optional parameters with default values for the offer building process as was requested here: #85 

**Misc**
I'm unsure which fields should be required and optional and why this is necessary in the first place. 
I'd rather argue, that we should avoid allowing offer creation without having all fields set, since it's a legal binding document without all the necesary information at hand.


Steffengreiner:
I reduced the necessary amount of parameters for the Offer DTO to the most basic requirements:
`offerId` to ensure each created Offer is unique and can be tracked
`projectManager` to ensure accountability for a created Offer.  
`modificationDate` to track when the Offer was created
Let me know if you think this is too little or if other fields are more paramount to be set as required

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimate.groovy
```diff
@@ -0,0 +1,147 @@
...
+        Customer customer
+        ProjectManager projectManager
+        String projectTitle
+        String projectDescription
+        List<ProductItem> items
+        double totalPrice
+        CostEstimateId identifier
+        Affiliation selectedCustomerAffiliation
+
+        Builder(Date modificationDate) {
```
jenniferboedker, line 72: I think the modification date is not really relevant if you think about transferring offer information from the view to the use case. I would suggest having the project title, project description, customer, customer affiliation, PM, and items

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -67,21 +68,53 @@ class Offer {
         OfferId identifier
         Affiliation selectedCustomerAffiliation
 
-        Builder(Date modificationDate, Date expirationDate, Customer customer, ProjectManager projectManager, String projectDescription, String projectTitle, List<ProductItem> items, double totalPrice, OfferId identifier, Affiliation selectedCustomerAffiliation) {
-            this.modificationDate = modificationDate
+        Builder(Date modificationDate) {
```
jenniferboedker, line 71: Same as for cost estimate

jenniferboedker (changes requested):
I would argue that we do need at least the minimal arguments to create an offer (as listed in one of my comments). The ID is something created by the use case and is added later to the information of the offer.

Other than that, thanks for taking care of the issue that I opened 👍

src/main/groovy/life/qbic/datamodel/dtos/business/CostEstimate.groovy
Steffengreiner: Alright I can do that 👍

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -28,7 +28,7 @@ class Offer {
     /**
      * The QBiC project manager who was assigned to the project
      */
-    final Person projectManager
+    final ProjectManager projectManager
```
jenniferboedker, line 31: 🚀

jenniferboedker (approved):
LGTM

PR #98, opened 2020-11-16
Add correct Product class to ProductItem

jenniferboedker:
This PR adds the new Product class to the ProductItem and adds the deprecated status to the old Product class


jenniferboedker:
We agreed in a meeting some time ago that ProductItmes do not require a description and a name, therefore they have been removed from this class.

src/test/groovy/life/qbic/datamodel/dtos/business/ProductItemSpec.groovy
```diff
@@ -1,8 +1,11 @@
 package life.qbic.datamodel.dtos.business
 
-import life.qbic.datamodel.accounting.Product
-import life.qbic.datamodel.accounting.ProductCategory
+import life.qbic.datamodel.dtos.business.services.AtomicProduct
+import life.qbic.datamodel.dtos.business.services.DataStorage
```
Steffengreiner, line 4:
```suggestion
```
I don't think these imports are necessary for your test.

src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
```diff
@@ -1,6 +1,7 @@
 package life.qbic.datamodel.accounting
 
 import groovy.transform.EqualsAndHashCode
+import life.qbic.datamodel.dtos.business.services.Product
 
 /**
  * An item with associated costs and quantity
```
Steffengreiner, line 7:
```suggestion
 * A ProductItem with associated costs and quantity
```

src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
```diff
@@ -24,26 +25,8 @@ class ProductItem {
      */
```
Steffengreiner, line 25: Remove "which" from the description line above

src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
Steffengreiner: The rest of the description needs to be adapted to, replace Cost items with ProductItem in line 9 e.g. and remove the @author tag
(I can't make a suggestion since it's part of the changed code :( )

Steffengreiner (dismissed):
Very Nice! I have some suggestions concerning the description, but since they don't impact the superb logic displayed here i'll not gatekeep your PR 👍

src/main/groovy/life/qbic/datamodel/accounting/ProductItem.groovy
jenniferboedker: I removed the cost item, but I think we can keep item since it is a short form of ProductItem

Steffengreiner (approved):
LGTM

PR #100, opened 2020-11-17
Make quantity modifiable

jenniferboedker:
The quantity of a ProductItem is something that needs to change by the request of the user. Therefore, the attribute should be no longer final


Steffengreiner (approved):
LGTM!

KochTobi:
This should not be changeable in a DTO later. There the value has to be extracted after transfer and packed into a new DTO before transfer.

Issue #105, opened 2020-11-19
Customer hashCode() not working as expected


KochTobi:
The hashCode for different customers is equal because all fields are in the super class.

![image](https://user-images.githubusercontent.com/11536497/99675157-2aa71580-2a77-11eb-8c66-eadc697ba428.png)

PR #107, opened 2020-11-19
Fix/#105

KochTobi:
Solves #105

### DIFF
--- a/src/test/groovy/life/qbic/datamodel/dtos/business/ProjectManagerSpec.groovy
+++ b/src/test/groovy/life/qbic/datamodel/dtos/business/ProjectManagerSpec.groovy
@@ -41,4 +41,14 @@ class ProjectManagerSpec extends Specification{
         assert projectManager.affiliations.isEmpty()
     }
 
+    def "different project managers result in different objects"() {
+        when: "two different project managers are created"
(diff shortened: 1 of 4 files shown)


src/test/groovy/life/qbic/datamodel/dtos/business/ProjectManagerSpec.groovy
```diff
@@ -41,4 +41,14 @@ class ProjectManagerSpec extends Specification{
...
+    def "different project managers result in different objects"() {
+        when: "two different project managers are created"
+        ProjectManager projectManager1 = new ProjectManager.Builder("Andrew", "PM", "a.p@d.ef").build()
+        ProjectManager projectManager2 = new ProjectManager.Builder("Betty", "PM", "b.p@d.ef").build()
+
+        then: "the project managers and their hash code are not equal"
+        projectManager1.hashCode() != projectManager2.hashCode()
+        projectManager1 != projectManager2
+    }
+
```
jenniferboedker, line 53: Maybe also adding the @equalsandhash. method to the project manager would be nice since you also added the test here

jenniferboedker (dismissed):
LGTM, nice work to catch that but

see my comment about how to handle the ProjectManager the same as the Customer.

jenniferboedker (approved):
Thanks 😇 

LGTM

PR #109, opened 2020-11-20
Release/1.12.0

Steffengreiner:
This release introduces an abstract builder pattern for the business DTOs of the QBiC business domain and adds support for the QBiC Template Sync and Github Actions


qube.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 1.12.0-SNAPSHOT
+current_version = 1.12.0
```
jenniferboedker, line 2: did you use qube to bump the version?

jenniferboedker (commented):
So depending on if you used qube for the bump of the version I would either approve or request changes ;)

qube.cfg
Steffengreiner: Yes i used the qube `bump-version` command with the qube 2.5.1 version to increase the versioning of `data-model-lib`
Interesting that it skipped to do so in the `qube.cfg `Nice Catch 👍

jenniferboedker: No its fine there is no problem. I was just curious in general.

jenniferboedker (approved):
LGTM

PR #126, opened 2020-12-11
Fix offer product items

jenniferboedker:
**Purpose of change**
Offer ProductItems always returned an empty list

**Changes**
Adjust the builder to accept the method parameter and add items to one of the tests to reflect the required items


jenniferboedker:
fixes #142

KochTobi (commented):
Hey, @jenniferboedker good catch. 
However, I would suggest to leave it optional and include a `items(List<ProductItem> items` method in the builder and removing the parameter from the Builder constructor.

KochTobi (approved):
Nice work! LGTM

PR #127, opened 2020-12-11
add price attributes to offer dto

jenniferboedker:
**Purpose of change**
The offer DTO should reflect all parts of a total offer price which includes the value-added taxes, overheads, and the net price

**Changes**
The offer DTO has three new attributes: VAT, overhead, and net. The tests for offer dtos has been extended.


sven1103 (approved):
lgtm!

PR #129, opened 2020-12-14
add equals and hash to products

jenniferboedker:
**Changes**
Add the equals and hash annotation to compare products


Steffengreiner (approved):
LGTM

PR #131, opened 2020-12-17
add currency for products

jenniferboedker:
**Purpose of change**
Products do not only have prices but also a currency for them

**Changes**
Products have now a currency attribute that is on default in Euro. Furthermore, a test for the currency is added to ensure that is value is in Euro.


jenniferboedker:
[Add currency to the unit price in the Overview](https://trello.com/c/WHCf3Kqs/419-add-currency-to-the-unit-price-in-the-overview)

Steffengreiner (approved):
LGTM

PR #140, opened 2021-01-15
Implement toString() for tomato-id

sven1103:
This PR overrides the toString() method of the tomatoId class and
implements a meaningful String representation of the identifier.


jenniferboedker (approved):
Nice work, thanks!

Issue #145, opened 2021-01-29
Report Generation in GitHub Actions fails on development branch


Steffengreiner:
**Describe the bug**
The Problem started with the [commit ](https://github.com/qbicsoftware/data-model-lib/commit/121c92ad452a3950b732fc29bc196fce665aee42) of the Template Branch into the development branch. 
Since then the report generation fails for each new commit. 

**Expected behavior**
The report generation github action should not fail and the maven site generation should include the generated documentation.

**How to Fix**
Add the necessary plugins and dependencies to the `pom.xml` of data-model-lib. This includes: 

As a dependency: 

- groovy-all (for GmavenPlus to run properly)

As a plugin repository: 
The URL to our Nexus(otherwise the groovydoc-maven-plugin won't be found) 
https://qbic-repo.qbic.uni-tuebingen.de/repository/maven-releases

To the build section: 

- maven-site-plugin (For site generation)
- maven-project-info-reports-plugin (For report generation) 
- groovydoc-maven-plugin (To include the generated reports into the project summary)

To the reporting section:

- groovydoc-maven-plugin (To include the generated reports into the project summary)

**Additional context**
This is the same issue as [#124](https://github.com/qbicsoftware/offer-manager-2-portlet/issues/124) in the offer-manager-2-portlet. 
There is also a PR ready to fix the issue in offer-manager-2-portlet [#279](https://github.com/qbicsoftware/offer-manager-2-portlet/pull/279)

jenniferboedker:
Can this be closed?

Steffengreiner:
yes i think so

PR #147, opened 2021-02-01
Include and adapt dependencies to allow report-generation

Steffengreiner:
Description of changes
This PR intends to solve the issue #145 in context of the groovydoc report generation:

1. Add the `Maven Site`plugin and `maven-project-info-reports-plugin` to the `pom.xml` to ensure that the build runs in Github Actions.
2. Adding the `groovy-all` dependency with a set versioning since GMavenPlus needs the an explicit setting of groovy.
3 Enable the correct groovydoc generation during the site generation by adding the `site execution` setting to `GMavenPlus plugin` in the parent module `pom.xml` .
4 Additionally the `GroovydocOutput` property has to set for the GmavenPlus Plugin, so it's generated in the target/site directory
5. Adding the `groovydoc-maven-plugin` to the reporting section of the `pom.xml `to enable the linkage from the project summary page to the generated gapidocs.
6. Include the current Nexus to the `plugin repository` section of the `pom.xml.` Otherwise maven won't know where to look for the `groovydoc-maven-plugin`

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,6 +21,25 @@
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
     </properties>
 
+    <pluginRepositories>
+        <pluginRepository>
+            <id>maven-central-plugins</id>
+            <url>
+                https://repo.maven.apache.org/maven2
+            </url>
+        </pluginRepository>
+        <pluginRepository>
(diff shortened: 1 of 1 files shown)


KochTobi (approved):
LGTM

PR #152, opened 2021-02-10
Decrease groovy version to 2.5.10 to avoid breaking changes

KochTobi:
Classes compiled with Groovy 3 are not compatible with groovy 2.5.10. Therefore `1.13.0` must still use Groovy `2.5.X`


jenniferboedker (approved):
lgtm

PR #160, opened 2021-02-17
Add List of ProductItems and their Net Prices associated with Overhead Pricing to Offer DTO 

Steffengreiner:
**Description of changes**
This PR Introduces 4 new Fields to the Offer DTO which are necessary fot the final Offer document. 

- List of ProductItems which are relevant for overhead calculation 
- List of ProductItems which are irrelevant for overhead calculation 
- Net value of ProductItems which are relevant for overhead calculation 
- Net value of ProductItems which are irrevelant for overhead calculation 

It also updates the Offer DTO Testing to account for the new fields. 

**Additional Information** 
Currently the final Offer just lists the ProductItems one after the other. The new version should be able to distinct the provided services into ProductItems with an associated Overhead cost and ProductItems without an associated Overhead cost. Therefore for both cases the netSum and ProductItem List should be transferable from the Offer entitity to the PDF Conversion

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
@@ -65,6 +65,14 @@ class Offer {
      */
     final Affiliation selectedCustomerAffiliation
 
+    final List<ProductItem> itemsWithOverhead
+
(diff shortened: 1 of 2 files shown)


jenniferboedker (approved):
LVGTM, nice work :)

PR #161, opened 2021-02-22
Rename offer description to objective

jenniferboedker:
Renames the offer description to objective based on issue [#262](https://github.com/qbicsoftware/offer-manager-2-portlet/issues/262)

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
@@ -35,7 +35,12 @@ class Offer {
     /**
      * A short description of the project
      */
+    @Deprecated
     final String projectDescription
+    /**
+     * A short objective of the project
+     */
+    final String projectObjective
     /**
      * A list of items for which the customer will be charged
      */
@@ -88,6 +93,7 @@ class Offer {
          */
         String projectTitle
         String projectDescription
+        String projectObjective
         Customer customer
         ProjectManager projectManager
         Affiliation selectedCustomerAffiliation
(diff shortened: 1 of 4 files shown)


jenniferboedker:
[Rename the project description to project objective (#262)](https://trello.com/c/n4meZGNZ/534-rename-the-project-description-to-project-objective-262)

jenniferboedker:
When simply renaming description to objective it must be considered that we do not desire the release of a new version of the lib. The simple renaming would require that since it would break our API. Therefore I reviewed the possibilities of how to introduce the new field without breaking functionality. I came up with the following possibilities:

  1. Make the objective field optional in order to not break the current Builder 
  2. Set the value of the objective field from the constructor also for the description field in order to maintain the field and mark it as deprecated (the implemented option)

In a discussion with @KochTobi we found that the last option is the best solution since we also make the objective field required (which would not be the case for the first option) and make it possible to simply remove the description field when required without a bigger refactoring required. It also makes it easy to use the current constructor as is and using the new variable.

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -111,21 +117,38 @@ class Offer {
...
+        String projectDescription
+
+        Builder(Customer customer, ProjectManager projectManager, String projectTitle, String projectObjective, Affiliation selectedCustomerAffiliation) {
+            /*
+            Overall offer describing properties
+             */
             this.customer = Objects.requireNonNull(customer, "Customer must not be null")
             this.projectManager = Objects.requireNonNull(projectManager, "Project Manager must not be null")
             this.projectTitle = Objects.requireNonNull(projectTitle, "Project Title must not be null")
-            this.projectDescription = Objects.requireNonNull(projectDescription, "Project Description must not be null")
```
sven1103, line 118: You need to set the description to an empty String, otherwise implementations will throw NullPointers :)

sven1103 (changes requested):
A small request, that will prevent existing implementations to break

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
jenniferboedker: Can I also set it the same as the projectObjective?

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -111,21 +117,38 @@ class Offer {
...
+            this.totalPrice = 0
+            this.itemsWithOverhead = []
+            this.itemsWithoutOverhead = []
+            this.itemsWithOverheadNet = 0
+            this.itemsWithoutOverheadNet = 0
+
+            /*
+            Deprecated
+             */
+            this.projectDescription = Objects.requireNonNull(projectObjective, "Project Objective must not be null")
```
jenniferboedker, line 151: @sven1103 the description is set here

sven1103: ah, thanks

sven1103 (approved):

PR #163, opened 2021-02-25
Fix/location date format

KochTobi:
Use the ISO_8601_DATETIME_FORMAT instead of a custom minute resolution

``"yyyy-MM-dd'T'HH:mm'Z'"`` -> ``"yyyy-MM-dd'T'HH:mm:ss'Z'"``


sven1103 (dismissed):
Nice!

sven1103:
Though you need to change the base branch imo

KochTobi:
Yes there was some movement in the last days.

jenniferboedker:
@KochTobi if this is an important solution maybe it would be good to document the issue (in an issue) and mention the solution so we can delete this branch.

KochTobi:
We need to conform with ISO standards otherwise REST calls with this object are hell!

PR #167, opened 2021-03-05
Add DTOs for project registration and management

sven1103:
This PR adds some classes to support the implementation of the openBIS integration in the new offer manager app.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -10,6 +10,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 **Added**
 
 * ``ProjectIdentifier``, ``ProjectCode``, ``ProjectSpace`` and ``Project`` to describe QBiC projects
+* ``ProjectApplication`` to describe a project application for registration at QBiC's data management platform
 
 **Fixed**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,17 @@ Changelog
 
 This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
+2.3.0
+--------
+
+**Added**
+
+**Fixed**
+
+**Dependencies**
+
+**Deprecated**
+
 
 2.2.0 (2021-03-02)
 ------------------
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,8 +9,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
-* ``ProjectIdentifier``, ``ProjectCode``, ``ProjectSpace`` and ``Project`` to describe QBiC projects
-* ``ProjectApplication`` to describe a project application for registration at QBiC's data management platform
+* ``life.qbic.datamodel.dtos.projectmanagement.ProjectIdentifier``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectCode``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectSpace`` and ``life.qbic.datamodel.dtos.projectmanagement.Project`` to describe QBiC projects
+* ``life.qbic.datamodel.dtos.projectmanagement.ProjectApplication`` to describe a project application for registration at QBiC's data management platform
 
 **Fixed**
 
@@ -338,4 +338,3 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 **Dependencies**
 
 **Deprecated**
-
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
+* ``ProjectIdentifier``, ``ProjectCode``, ``ProjectSpace`` and ``Project`` to describe QBiC projects
+
 **Fixed**
 
 **Dependencies**
--- a/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
@@ -8,7 +8,7 @@ import life.qbic.datamodel.dtos.business.OfferId
  * Describes and provides information about a scientific project
  * at QBiC.
  *
- * @since 1.0.0
+ * @since 2.3.0
  */
 class Project {
 
--- a/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
@@ -1,4 +1,6 @@
-package life.qbic.datamodel.dtos.business
+package life.qbic.datamodel.dtos.projectmanagement
+
+import life.qbic.datamodel.dtos.business.OfferId
 
 /**
  * A typical scientific QBiC project.
--- a/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
@@ -52,7 +52,7 @@ class ProjectApplication {
     /**
      * The requested project space the project shall be associated with (optional)
      */
-    final Optional<ProjectSpace> projectSpace
+    ProjectSpace projectSpace
 
     /**
      * The desired project code (optional)
@@ -67,7 +67,7 @@ class ProjectApplication {
         String experimentalDesign
         ProjectManager projectManager
         Customer customer
-        Optional<ProjectSpace> projectSpace
+        ProjectSpace projectSpace
         Optional<ProjectCode> projectCode
 
         Builder(OfferId linkedOffer,
@@ -75,6 +75,7 @@ class ProjectApplication {
                 String projectObjective,
                 String experimentalDesign,
                 ProjectManager projectManager,
+                ProjectSpace projectSpace,
                 Customer customer) {
             this.linkedOffer = Objects.requireNonNull(linkedOffer)
             this.projectTitle =  Objects.requireNonNull(projectTitle)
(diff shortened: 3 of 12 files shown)


sven1103:
supports https://github.com/qbicsoftware/offer-manager-2-portlet/pull/355

sven1103:
[Create new Objects in the data model lib that support the new offer manager functionality for the openBIS use case.](https://trello.com/c/5KyfkOTr/579-create-new-objects-in-the-data-model-lib-that-support-the-new-offer-manager-functionality-for-the-openbis-use-case)

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
```diff
@@ -0,0 +1,35 @@
...
+ * at QBiC.
+ *
+ * @since 1.0.0
+ */
+class Project {
+
+    /**
+     * A short but descriptive project title
+     */
+    final String projectTitle
```
jenniferboedker, line 18: Do we need a project title? How is it different from the title of the project title of the offer?

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
```diff
@@ -0,0 +1,35 @@
...
+
+import life.qbic.datamodel.dtos.business.OfferId
+
+/**
+ * A typical scientific QBiC project.
+ *
+ * Describes and provides information about a scientific project
+ * at QBiC.
+ *
+ * @since 1.0.0
```
jenniferboedker, line 11:
```suggestion
 * @since 2.3.0
```

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
```diff
@@ -0,0 +1,114 @@
...
+ *
+ * Beside the mandatory fields required to be passed to the constructor, there are two optional
+ * fields:
+ *   1: projectSpace
+ *   2: projectCode
+ * If any of these are not defined, then either of it will be created randomly.
+ *
+ * @since 2.3.0
+ */
+class ProjectApplication {
```
jenniferboedker, line 20:
```suggestion
class ProjectCandidate {
```

jenniferboedker: When I hear `application` I think of it in the context of an app

CHANGELOG.rst
```diff
@@ -4,6 +4,20 @@ Changelog
 
 This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
+2.3.0
+--------
+
+**Added**
+
+* ``ProjectIdentifier``, ``ProjectCode``, ``ProjectSpace`` and ``Project`` to describe QBiC projects
+* ``ProjectApplication`` to describe a project application for registration at QBiC's data management platform
```
KochTobi, line 13: Please also provide the complete identifier since this can lead to problems later (we have multiple Person classes for example).
```suggestion
* ``life.qbic.datamodel.dtos.projectmanagement.ProjectIdentifier``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectCode``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectSpace`` and ``life.qbic.datamodel.dtos.projectmanagementProject`` to describe QBiC projects
* ``life.qbic.datamodel.dtos.projectmanagementProjectApplication`` to describe a project application for registration at QBiC's data management platform
```

```suggestion
* ``ProjectIdentifier``, ``ProjectCode``, ``ProjectSpace`` and ``Project`` to describe QBiC projects in ``life.qbic.datamodel.dtos.projectmanagement``
* ``ProjectApplication`` to describe a project application for registration at QBiC's data management platform in ``life.qbic.datamodel.dtos.projectmanagement``
```

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Information about a desired project registration in QBiC's project management system.
+ *
+ * Instances of this class describe a request to register a new project at QBiC.
+ *
+ * Beside the mandatory fields required to be passed to the constructor, there are two optional
+ * fields:
+ *   1: projectSpace
+ *   2: projectCode
+ * If any of these are not defined, then either of it will be created randomly.
```
jenniferboedker, line 16: Shouldn't they be created by the system it uses? I would not see them as optional.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * The associated offer
+     */
+    final OfferId linkedOffer
+
+    /**
+     * A short and precise project title
+     */
+    final String projectTitle
```
jenniferboedker, line 30: Again, I am not sure what information the title here can have what we do not already have in our offer

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * The associated project manager
+     */
+    final ProjectManager projectManager
+
+    /**
+     * The associated customer
+     */
+    final Customer customer
```
jenniferboedker, line 50: same as for the title

KochTobi:
`PR to master branch from patch/release branch only / test (pull_request)` fails

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * The requested project space the project shall be associated with (optional)
+     */
+    final Optional<ProjectSpace> projectSpace
+
+    /**
+     * The desired project code (optional)
+     */
+    final Optional<ProjectCode> projectCode
```
jenniferboedker, line 60: I think they should not be optional

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectCode.groovy
```diff
@@ -0,0 +1,39 @@
+package life.qbic.datamodel.dtos.projectmanagement
+
+/**
+ * Describes the project code, that identifies the project within a project space.
+ *
+ * A project code is the unique identifier of a QBiC project within a project space.
+ *
+ * A project code can be validated with the following regular expression: "Q[A-X0-9]{4}".
+ *
+ * @since 1.0.0
```
jenniferboedker, line 10:
```suggestion
 * @since 2.3.0
```

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectCode.groovy
```diff
@@ -0,0 +1,39 @@
...
+ */
+class ProjectCode {
+
+    final String code
+
+    private static final def REGEX = ~'Q[A-X0-9]{4}'
+
+    /**
+     * Constructs a project code instance based on the given code String.
```
jenniferboedker, line 19:
```suggestion
     * Constructs a project code instance based on the given code string.
```

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
```diff
@@ -0,0 +1,47 @@
...
+ *  4. Trailing and leading white space are trimmed
+ *
+ * @since 2.3.0
+ */
+final class ProjectSpace {
+
+    /**
+     * The project space name
+     */
+    final String name
```
jenniferboedker, line 22: Is there any rule on the name. What does it contain? E.g Customer name + project name or so?

jenniferboedker (changes requested):
Nice work!
I added some questions where in general it was unclear to me which information the DTOs really need to contain

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
sven1103: it is not, but it is a context change we have here. we left the offer realm and are in the project life cycle now. I don't want to pollute contexts by mixing up objects, just to provide the information. I want it very declarative.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
sven1103: No, the pm can leave them optional or set them explicitly if required.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
sven1103: I disagree, you apply for an project registration, and application is the right word describing it.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
sven1103: same as above, context!

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectApplication.groovy
sven1103: yes, they can be :) if the code is not set, one will be provided then.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
sven1103: Rule see the format space name and class description. There is an outdated "rule" the project managers use, but as we discussed in the stakeholder interview, this was merely a hack, because they were not able to filter otherwise. we have the properties available in the offer (the selected customer affiliation).

jenniferboedker:
In a short discussion, we agreed that the space is not optional anymore but should be given by the user since it can be arbitrary. Sample codes can be left optional since they have a pattern based on which they can be created.

jenniferboedker:
I have another question, why do we distinguish between `Project` and `ProjectApplication` on the entity level? Is it not sufficient to use only the `Project` object? In essence, they do contain the same information. Only one project is registered and the other one is not e.g. we also didn't do that for offers.

sven1103:
> I have another question, why do we distinguish between `Project` and `ProjectApplication` on the entity level? Is it not sufficient to use only the `Project` object? In essence, they do contain the same information. Only one project is registered and the other one is not e.g. we also didn't do that for offers.

Yes, i want to keep the option open to extend the project object with additional properties, but leave the project application unaffected. SRP.

jenniferboedker (approved):

PR #168, opened 2021-03-08
Introduce OverheadRatio Property in Offer DTO 

Steffengreiner:
- [X] This comment contains a description of changes (with reason)
- [X] Referenced issue is linked
- [X] If you've fixed a bug or added code that should be tested, add tests!
- [Never!] Documentation in `docs` is updated
- [X] `CHANGELOG.rst` is updated

**Description of changes**
It was requested by the stakeholders to show the overhead percentage in the final Offer PDF. (See [#358](https://github.com/qbicsoftware/offer-manager-2-portlet/issues/358)) 
For this reason this PR introduces the overheadRatio property to the Offer DTO. 
The intent of this property is to store the applicable overhead markup, which is dependent on the customer affiliation.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -11,20 +11,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 * ``life.qbic.datamodel.dtos.projectmanagement.ProjectIdentifier``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectCode``, ``life.qbic.datamodel.dtos.projectmanagement.ProjectSpace`` and ``life.qbic.datamodel.dtos.projectmanagement.Project`` to describe QBiC projects
 * ``life.qbic.datamodel.dtos.projectmanagement.ProjectApplication`` to describe a project application for registration at QBiC's data management platform
-
-**Fixed**
-
-**Dependencies**
-
-**Deprecated**
-
-
-2.3.0-SNAPSHOT (2021-03-05)
----------------------------
-
-**Added**
-
-* overheadModifier property for ``life.qbic.datamodel.dtos.business.Offer``
+* ``life.qbic.datamodel.dtos.business.Offer#overheadModifier` property for ``life.qbic.datamodel.dtos.business.Offer```
 
 **Fixed**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,14 +4,15 @@ Changelog
 
 This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
-2.3.0
---------
+
+2.3.0-SNAPSHOT (2021-03-05)
+---------------------------
 
 **Added**
 
+* overheadModifier property for ``life.qbic.datamodel.dtos.business.Offer``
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -24,6 +24,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
+* overheadModifier property for ``life.qbic.datamodel.dtos.business.Offer``
+
```
KochTobi, line 28: Why is there the `2.3.0-SNAPSHOT` and the `2.3.0` What happened here ??

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -91,6 +91,10 @@ class Offer {
      * The net price of all items for which an overhead cost is not applicable, without overhead and taxes
      */
     final double itemsWithoutOverheadNetPrice
+    /**
+     * The overhead modifier applied to the pricing dependent on the customer affiliation
+     */
+    final double overheadModifier
```
KochTobi, line 97: Is this really offer specific? Isn't it dependent on the affiliation?

KochTobi (commented):
Hi @Steffengreiner this looks good however, it is unclear to me why you need this change and whether it is logically correct or not.

CHANGELOG.rst
Steffengreiner: The `2.3.0-Snapshot` section was added to the changelog during the qube force bump. I'm unsure why it's in the changelog. I can remove it and move the overheadModifier property mention to the 2.3.0 section 👍

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
Steffengreiner: The applicable overhead modifier is dependent on the affiliation stored in the Offer. 
I could also add a small method to the OfferPDF Conversion that extracts the affilation and then assigns the overhead cost. However we already have a method in the Offer Entity in the Offer-manager that determines the overhead modifier so i thought it would be better to avoid duplicate methods.

CHANGELOG.rst
KochTobi: I think the 2.3.0-Snapshot is correct but the 2.3.0 is wrong

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
KochTobi: I think it makes more sense to add the method that extracts that property since it does not belong to the offer but the affiliation associated with the customer in the offer.

Steffengreiner: But then that means as soon as the overhead cost or affiliation change you would have to update it seperately in the Offer2PDF converter class and in the Offer entitity

CHANGELOG.rst
```diff
@@ -11,20 +11,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
...
-**Deprecated**
-
-
-2.3.0-SNAPSHOT (2021-03-05)
----------------------------
-
-**Added**
-
-* overheadModifier property for ``life.qbic.datamodel.dtos.business.Offer``
+* ``life.qbic.datamodel.dtos.business.Offer#overheadModifier` property for ``life.qbic.datamodel.dtos.business.Offer```
```
KochTobi, line 14: shouldn't it be the other way around? e.g. everything is in the 2.3.0-snapshot section?

KochTobi (approved):

PR #169, opened 2021-03-09
Fix ProjectApplication Builder

jenniferboedker:
The creation of the ProjectAppliction did not function.
Reasons for this were that the Builder was not static and that the ProjectCode was still static in the builder but not in the class


KochTobi:
Why is the project code now mandatory?

KochTobi (approved):

jenniferboedker:
[Implement the domain logic for the Create Project and Create Project Space use case](https://trello.com/c/YCcQtA9N/576-implement-the-domain-logic-for-the-create-project-and-create-project-space-use-case)

PR #170, opened 2021-03-10
Replace dashes with underscore for project space names

sven1103:


src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
```diff
@@ -1,17 +1,19 @@
...
- *  4. Trailing and leading white space are trimmed
+ * <p>A project space has a name that follows this convention:</p>
  *
+ * <ul>
+ *  <li> No white space </li>
+ *  <li> Capital letters only </li>
+ *  <li> Inner white space is replaced by `_` </li>
+ *  <li> Trailing and leading white space are trimmed </li>
+ *  <li> Dashes are replaced by `_` </li>
+ * </ul>
```
jenniferboedker, line 16: 🚀

jenniferboedker, line 16: ul is the unordered list and results in bullet points, if you want to have an ordered list just as before you need to use ol
```suggestion
 * <ol>
 *  <li> No white space </li>
 *  <li> Capital letters only </li>
 *  <li> Inner white space is replaced by `_` </li>
 *  <li> Trailing and leading white space are trimmed </li>
 *  <li> Dashes are replaced by `_` </li>
 * </ol>
```

jenniferboedker (approved):
looks good 👍🏼

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
sven1103: thank you @jenniferboedker . Yes i figured out that the order is not necessary, so i went for ul.

PR #171, opened 2021-03-12
Move ProjectApplication from projectmanagement to business

KochTobi:

## Description of Changes
The ProjectApplication class was moved from projectmanagement to business. It is only applicable in the context of offers and therefore belongs in this package.


Steffengreiner (approved):
LGTM!

PR #172, opened 2021-03-12
Introduce Builder and make offerId optional

KochTobi:
Removes the necessity to provide an OfferId to a Project and introduces a Builder.


Steffengreiner (approved):
LGTM!

PR #173, opened 2021-03-15
Feature/adjust product

KochTobi:
The product now provides the version e.g. unique id as long instead of a string.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/ProductId.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/ProductId.groovy
@@ -56,7 +56,7 @@ class ProductId {
             return this
         }
 
-        Builder identifier(long identifier) {
+        Builder uniqueId(long identifier) {
             this.uniqueId = identifier
             return this
         }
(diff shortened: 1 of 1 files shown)


src/main/groovy/life/qbic/datamodel/dtos/business/ProductId.groovy
```diff
@@ -23,27 +23,86 @@ class ProductId {
...
+    static class Builder {
+        private String productType
+        private long uniqueId
+
+        /**
+         *
+         * @param productType
+         * @param identifier the unique id - will be interpreted as unsigned long
+         */
+        Builder(String productType, String identifier) {
```
jenniferboedker, line 40: I think the Typ is also fixed for the different products (or should be) like DS for data storage.

KochTobi: Shall I change it? Do we haven an enum already? I have the impression that we use the distinction rather often. And I would suggest to implement that in the classes to get their product type string

jenniferboedker (approved):

PR #182, opened 2021-03-18
Fix equality check for service products

sven1103:
Overrides the equals(Object obj) method for the `ProductId`.


KochTobi (dismissed):
LGTM ! Nice tests

KochTobi (dismissed):
Even better with changelog entry.

KochTobi (approved):

PR #185, opened 2021-03-19
Add product constructor with running number as long

jenniferboedker:
**Purpose**
The product classes still used the old constructor for building the product id, the new builder pattern with the running number as a long should be used.

**Changes**
A new constructor is added for each class. The old constructor is deprecated and the changes are logged in the CL


sven1103 (approved):
Well done and nice spot!

PR #189, opened 2021-03-23
Remove deprectation from sampel tracking classes

KochTobi:
The classes cannot be replaced by any other class in the data-model-lib. Thus deprecation can be removed and was wrong in the first place. (I did it, sorry)

- [x] This comment contains a description of changes (with reason)


Steffengreiner (approved):
LGTM!

PR #191, opened 2021-03-29
Feature/add unknown product category

KochTobi:
Many thanks for contributing to this project!
Addresses https://github.com/qbicsoftware/offer-manager-2-portlet/issues/443

**PR Checklist**
Please fill in the appropriate checklist below (delete whatever is not relevant). These are the most common things requested on pull requests (PRs).

- [x] This comment contains a description of changes (with reason)
- [x] Referenced issue is linked

**Description of changes**
There currently is no default value for the product category. For applications that might lead to errors in case the database contains values that are not reflected in the ProductCategory enum. For those cases, I introduce an `UNKNOWN` category that should only be used in case the category received is not reflected by the other enum values. Is this the case, then the applications have to make sure to handle products with unknown category correctly.


KochTobi:
The decision was made that an unknown category should not be included. @jenniferboedker @Steffengreiner

PR #197, opened 2021-05-06
Feature/update pipeline schema

sven1103:
This PR updates the JSON schema for bioinformatic analysis run output.

The additions address the two additional files we need to check, which are the

- input_ids.txt
- run_id.txt

files. The `input_ids.txt` will hold a list of sample ids, that reference the datasets that have been used for the analysis. The `run_id` contains the unique Nextflow Tower run id that identifies the pipeline run.

### DIFF
--- a/src/test/resources/examples/resultset/valid-resultset-example.json
+++ b/src/test/resources/examples/resultset/valid-resultset-example.json
@@ -1,7 +1,7 @@
 {
   "pipelineInformation": {
     "name": "pipeline_info",
-    "path": "./",
+    "path": "./pipeline_info",
     "children": [],
     "softwareVersions": {
       "name": "software_versions",
@@ -21,7 +21,7 @@
   },
   "qualityControl": {
     "name": "multiqc",
-    "path": "./",
+    "path": "./multiqc",
     "children": [
       {
         "name": "star_salmon",
@@ -39,7 +39,7 @@
   "processFolders": [
     {
       "name": "salmon",
-      "path": "./",
+      "path": "./salmon",
       "children": [
         {
           "name": "salmon.merged.gene_tpm",
--- a/src/test/resources/examples/resultset/valid-resultset-example.json
+++ b/src/test/resources/examples/resultset/valid-resultset-example.json
@@ -6,17 +6,17 @@
     "softwareVersions": {
       "name": "software_versions",
       "fileType": "csv",
-      "path": "pipeline_info/software_versions"
+      "path": "./pipeline_info/software_versions.csv"
     },
     "pipelineReport": {
       "name": "pipeline_report",
       "fileType": "txt",
-      "path": "pipeline_info/pipeline_report"
+      "path": "./pipeline_info/pipeline_report.txt"
     },
     "executionReport": {
       "name": "execution_report",
       "fileType": "txt",
-      "path": "pipeline_info/execution_report"
+      "path": "./pipeline_info/execution_report.txt"
     }
   },
   "qualityControl": {
@@ -25,11 +25,11 @@
     "children": [
       {
         "name": "star_salmon",
-        "path": "multiqc/star_salmon",
+        "path": "./multiqc/star_salmon",
         "children": [
           {
             "name": "multiqc_report.html",
-            "path": "multiqc/star_salmon/multiqc_report.html",
+            "path": "./multiqc/star_salmon/multiqc_report.html",
             "fileType": "html"
           }
         ]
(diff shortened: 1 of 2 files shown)


Steffengreiner (approved):
LGTM! 👍 Don't forget to update the ER diagram

KevinMenden (dismissed):
Nice, looks good!

Steffengreiner (dismissed):
Very nice!

ggabernet:
Is this the general schema for all pipeline results registration? Because there are some files that refer to `salmon` and these would be specific for the RNAseq pipeline. Maybe we should get rid of those

src/test/resources/examples/resultset/valid-resultset-example.json
```diff
@@ -39,14 +39,24 @@
...
+  ],
+  "runID": {
+    "name": "run_id",
+    "fileType": "txt",
+    "path": "./run_id.txt"
+  },
+  "inputIDs": {
+    "name": "input_ids",
+    "fileType": "txt",
+    "path": "./input_ids.txt"
```
ggabernet, line 60: So that's the file containing the Sample IDs for qPostman, right?
I would maybe suggest to call this one `sampleIDs` so that it is clear that it refers to the sample IDs (QBIC code).

src/test/resources/examples/resultset/valid-resultset-example.json
```diff
@@ -39,14 +39,24 @@
...
           "name": "salmon.merged.gene_tpm",
           "fileType": "tsv",
-          "path": "salmon/salmon.merged.gene_tpm"
+          "path": "./salmon/salmon.merged.gene_tpm"
         }
       ]
     }
-  ]
+  ],
+  "runID": {
```
ggabernet, line 52: Regarding the `runID`, that is given by Tower, do you recall if we get returned this ID from Tower when calling the API, @KevinMenden ?
I'm just trying to think how we would programmatically get this one.

src/test/resources/examples/resultset/valid-resultset-example.json
```diff
@@ -1,35 +1,35 @@
...
     "children": [
       {
         "name": "star_salmon",
-        "path": "multiqc/star_salmon",
+        "path": "./multiqc/star_salmon",
         "children": [
           {
             "name": "multiqc_report.html",
-            "path": "multiqc/star_salmon/multiqc_report.html",
+            "path": "./multiqc/star_salmon/multiqc_report.html",
```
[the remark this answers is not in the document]
ggabernet: Even though it might not be so relevant for the example, the changes that I made reflects how most pipelines are structured. Subfolders there are a special case of the RNAseq pipelines and maybe a couple more.

src/test/resources/examples/resultset/valid-resultset-example.json
KevinMenden: Hmm I don't really remember how that worked, but I think you might even have to know that to access the API anyway :thinking: so we would have to store it somewhere anyway I guess.

sven1103:
> Is this the general schema for all pipeline results registration? Because there are some files that refer to `salmon` and these would be specific for the RNAseq pipeline. Maybe we should get rid of those

I think you are referring to the example json right? They are just for the tests, to see if the schema validation is behaving as intended.

src/test/resources/examples/resultset/valid-resultset-example.json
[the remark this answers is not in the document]
sven1103: true, but its a valid example, and that is what this json represents. In the schema we defined `qualityControl` as a `DataFolder`, so it can be arbitrarily nested with DataFiles and DataFolders. The only thing we can check for validation is, that it has `multiqc` in the name of the top level multiqc folder.

src/test/resources/examples/resultset/valid-resultset-example.json
sven1103: if the API follows to HATEOAS principle, than the HTTP response should return a URL to this resource (containing the run id).

src/test/resources/examples/resultset/valid-resultset-example.json
sven1103: addressed in ec2d1dc

src/test/resources/examples/resultset/valid-resultset-example.json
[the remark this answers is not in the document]
ggabernet: ok sure then it is fine to leave it like this, just to mention that that would be a more common example 👍

src/test/resources/examples/resultset/valid-resultset-example.json
ggabernet: Yes, exactly, the API response returns the Run ID, tested it yesterday :)

PR #210, opened 2021-05-17
Introduce Unspecified Files for nfCore Parser

Steffengreiner:
- [X] This comment contains a description of changes (with reason)
- [X] `CHANGELOG.rst` is updated

**Description of changes**
This PR introduces the `UnspecifiedFile` class which extends the `DataFile` Object. 
This class will be used to represent files which can't be accounted for in the schema since they are dependent on the pipeline(e.g. the files in the subfolders of the `ProcessFolder` class). 

**Additional context**
I'm unsure if the class name could not be improved into something like `UnspecifedNfCoreFile` or something simliar.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/datasets/datastructure/files/nfcore/UnspecifiedFile.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/datastructure/files/nfcore/UnspecifiedFile.groovy
@@ -1,0 +1,30 @@
+package life.qbic.datamodel.datasets.datastructure.files.nfcore
+
+import life.qbic.datamodel.datasets.datastructure.files.DataFile
+
+/**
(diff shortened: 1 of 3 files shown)


src/main/groovy/life/qbic/datamodel/datasets/datastructure/files/nfcore/UnspecifiedFile.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+/**
+ * A specialisation of a DataFile, represents a non-specified/variable output file in the nf-core Pipeline
+ *
+ * @since 2.7.0
+ */
+class UnspecifiedFile extends DataFile {
+
+    protected UnspecifiedFile() {}
+
```
jenniferboedker, line 13: Do we really want to allow all possible files? There could then also be files accepted that we do not want to have.

PR #211, opened 2021-05-17
Make ProductId comparable

jenniferboedker:
In order to properly sort the product identifier it must implement the comparable interface


src/main/groovy/life/qbic/datamodel/dtos/business/ProductId.groovy
```diff
@@ -124,4 +124,11 @@ class ProductId {
     String toString() {
         return "${type}_${uniqueId}"
     }
+
+    @Override
+    int compareTo(ProductId productId2) {
+        //if the types are equal compare by uniqueId
```
KochTobi, line 130: I would suggest using the equals method here to make sure it behaves as expected.
```suggestion
        //if the types are equal compare by uniqueId
        if(this.equals(productId2)) return 0
```

KochTobi (changes requested):
I would suggest using the equals method as well in the compareTo method or adding the note "Note: this class has a natural ordering that is inconsistent with equals." to the JavaDoc

src/main/groovy/life/qbic/datamodel/dtos/business/ProductId.groovy
jenniferboedker: done :)

KochTobi: Sorry but that is not what i meant :) 
The operator in the type comparison is fine. What I meant is, that the productId compareTo method needs to be in sync with the equals method of ProductId

jenniferboedker:
[Service Products: Default product sorting #574](https://trello.com/c/gaIZzYZN/716-service-products-default-product-sorting-574)

KochTobi (approved):
Looks good :+1:

PR #219, opened 2021-06-14
Add flow cell as product unit!

jenniferboedker:
This PR adds per flow cell as a product unit


jenniferboedker:
Will be release with patch release 2.8.1

PR #229, opened 2021-07-01
Add PortalUser

sven1103:
Adds a simple portal user DTO

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,7 +9,7 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
-* A new class ``life.qbic.datamodel.dtos.portal.PortalUser`` that represents a user within QBiC's data management platform.
+* A new class ``life.qbic.datamodel.dtos.portal.PortalUser`` that represents a user within QBiC's data management platform. (`#229 <https://github.com/qbicsoftware/data-model-lib/pull/229>`_)
 
 **Fixed**
 
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -9,6 +9,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
+* A new class ``life.qbic.datamodel.dtos.portal.PortalUser`` that represents a user within QBiC's data management platform.
+
 **Fixed**
 
 **Dependencies**
--- a/src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
@@ -1,0 +1,46 @@
+package life.qbic.datamodel.dtos.portal
+
+import groovy.transform.Immutable
+
+/**
+ * <p>A user within QBiC's data management platform.</p>
+ *
+ * <p>For example a user with the full name <strong>Max Maxwell Minestrone</strong> has
+ * the first name <i>Max</i>, the middle name <i>Maxwell</i> and the last name <i>Minestrone</i>.</p>
+ *
+ * @since 2.9.0
+ */
+@Immutable
+class PortalUser {
+
+    /**
+     * The portal system's user id
+     * @since 2.9.0
+     */
+    String id
+
+    /**
+     * The full name of the user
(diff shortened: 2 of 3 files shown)


CHANGELOG.rst
```diff
@@ -9,6 +9,8 @@ This project adheres to `Semantic Versioning <https://semver.org/>`_.
 
 **Added**
 
+* A new class ``life.qbic.datamodel.dtos.portal.PortalUser`` that represents a user within QBiC's data management platform.
```
jenniferboedker, line 12:
```suggestion
* A new class ``life.qbic.datamodel.dtos.portal.PortalUser`` that represents a user within QBiC's data management platform. (`#229 <https://github.com/qbicsoftware/data-model-lib/pull/229>`_)
```

src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
```diff
@@ -0,0 +1,46 @@
...
+     * The first name of the user
+     * @since 2.9.0
+     */
+    String firstName
+
+    /**
+     * The middle name of the user
+     * @since 2.9.0
+     */
+    String middleName
```
jenniferboedker, line 38: what happens if someone does not have a middle name?

jenniferboedker: isnt that Optional?

src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
```diff
@@ -0,0 +1,46 @@
...
+     * The portal system's user id
+     * @since 2.9.0
+     */
+    String id
+
+    /**
+     * The full name of the user
+     * @since 2.9.0
+     */
+    String fullName
```
jenniferboedker, line 26: This is kind of redundant information since you can retrieve that information from the first (, middle) and last name

jenniferboedker (changes requested):
Thanks for your contribution, I also like that you include new styles e.g the immutable. 
I was wondering if we really need a middle name and also if it might make sense to reuse the [person](https://github.com/qbicsoftware/data-model-lib/tree/master/src/main/groovy/life/qbic/datamodel/people) class here. We would need the email of the person later on anyway. 
Anyhow, when you want to keep as is we need to consider that not everyone has a middle name and with the immutable, it seems to me that the middle name is enforced for everyone.

jenniferboedker:
[Create class that holds logged-in user information](https://trello.com/c/1yEmluQm/827-create-class-that-holds-logged-in-user-information)

src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
sven1103: sure, then you just pass an empty string upon construction

src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
sven1103: true, i wanted to provide a getter for the full name, I will provide one

src/main/groovy/life/qbic/datamodel/dtos/portal/PortalUser.groovy
jenniferboedker: I would think it would mostly be like we provide the empty string which I personally dont like so much. In how many cases do we really have middle name?

src/test/groovy/life/qbic/datamodel/dtos/portal/PortalUserSpec.groovy
```diff
@@ -0,0 +1,25 @@
...
+        String lastName = "Fillinger"
+        String email = "sven.fillinger@qbic.uni-tuebingen.de"
+
+        when:
+        PortalUser user = new PortalUser.Builder(id, firstName, lastName, email).build()
+
+        then:
+        user.getUserId().equals("1234")
+    }
+
```
jenniferboedker, line 24: Niice 🚀

jenniferboedker (approved):
Looks good!

Issue #236, opened 2021-07-06
Project DTO is missing the equals method


KochTobi:
**Describe the bug**
[Project DTO](https://github.com/qbicsoftware/data-model-lib/blob/master/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy) is missing the equals method

**Expected behaviour**
2 DTOs with the same information are equal. This is not the case for projects atm.

KochTobi:
Closed by #240

PR #240, opened 2021-07-19
Add tests and implement equals method for Project and ProjectIdentifier DTO

sven1103:
Addresses #236


src/test/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpec.groovy
```diff
@@ -0,0 +1,54 @@
...
+
+        when:
+        boolean comparison = projectA.equals(projectB)
+
+        then:
+        assert comparison
+
+    }
+
+    def "Compare unequal projects shall evaluate to false"() {
```
KochTobi, line 37: This does only test the case when both attributes are different. It does not guarantee that projects with different titles (exclusive or) different ids are not equal.

src/test/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifierSpec.groovy
```diff
@@ -0,0 +1,36 @@
...
+        boolean comparison = idA.equals(idB)
+
+        then:
+        assert comparison
+    }
+
+    def "Compare to unequal ids"() {
+        given:
+        ProjectIdentifier idA = new ProjectIdentifier(new ProjectSpace("TEST"), new ProjectCode("QTEST"))
+        ProjectIdentifier idB = new ProjectIdentifier(new ProjectSpace("REMO"), new ProjectCode("QTEST"))
```
KochTobi, line 27: this does not test for different project codes leading to different product identifiers.

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifier.groovy
```diff
@@ -33,4 +33,17 @@ class ProjectIdentifier {
...
+    boolean equals(Object obj) {
+        if (this.is(obj)) {
+            return true
+        }
+        if (!obj instanceof ProjectIdentifier) {
+            return false
+        }
+        ProjectIdentifier otherIdentifier = (ProjectIdentifier) obj
+        return this.projectSpace.equals(otherIdentifier.getProjectSpace()) \
+                && this.projectCode.equals(otherIdentifier.getProjectCode())
```
KochTobi, line 47: You could circumvent the casting here.
```suggestion
        } else {
            return this.projectSpace.equals(obj.getProjectSpace()) \
                && this.projectCode.equals(obj.getProjectCode())
        }
```
In this case `obj` is always instance of `ProjectIdentifier`

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
```diff
@@ -10,6 +11,7 @@ import life.qbic.datamodel.dtos.business.OfferId
  *
  * @since 2.3.0
  */
+@EqualsAndHashCode
```
KochTobi, line 14: Do we need to include fields here?

KochTobi (changes requested):
Hi @sven1103 

I had a look at your changes. Thanks for tackling the problem. The code looks good. I noticed some test cases missing

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/Project.groovy
sven1103: no, since they are all public

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifier.groovy
sven1103: true, I guess I just prefer the explicit style of type conversion ;)

KochTobi (approved):
Nice work :+1:

PR #246, opened 2021-07-26
Add discount properties

jenniferboedker:
This PR adds the totalPrice and discountPrice to the productitem and the totalDiscount price to the offer DTO

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
@@ -73,6 +73,10 @@ class Offer {
      */
     final double overheads
     /**
+     * The sum of all discounts for this offer
+     */
+    final double totalDiscount
+    /**
      * The identifier for the offer which makes it distinguishable from other offers
      */
     final OfferId identifier
@@ -132,6 +136,7 @@ class Offer {
          */
         double totalPrice
         double netPrice
+        double totalDiscount
         double taxes
         double overheads
         double itemsWithOverheadNet
@@ -165,6 +170,7 @@ class Offer {
             this.overheads = 0
             this.taxes = 0
             this.totalPrice = 0
+            this.totalDiscount = 0
             this.itemsWithOverhead = []
             this.itemsWithoutOverhead = []
             this.itemsWithOverheadNet = 0
@@ -263,6 +269,11 @@ class Offer {
             return this
         }
 
+        Builder totalDiscount(double totalDiscount){
+            this.totalDiscount = totalDiscount
+            return this
+        }
+
         Offer build() {
             return new Offer(this)
         }
@@ -306,6 +317,7 @@ class Offer {
         this.itemsWithOverheadNetPrice = builder.itemsWithOverheadNet
         this.itemsWithoutOverheadNetPrice = builder.itemsWithoutOverheadNet
         this.overheadRatio = builder.overheadRatio
+        this.totalDiscount = builder.totalDiscount
 
         if (builder.associatedProject.isPresent()) {
             this.associatedProject = Optional.of(builder.associatedProject.get())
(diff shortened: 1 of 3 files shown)


jenniferboedker:
[#686: adjust DTOs in DML to represent the discount in the product items](https://trello.com/c/SpHnrSo1/893-686-adjust-dtos-in-dml-to-represent-the-discount-in-the-product-items)

src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
```diff
@@ -23,10 +23,39 @@ class ProductItem {
...
 
+    /**
+     * Describes the total price of an item based on the quantity and unit price of a product
+     */
+    final double totalPrice
+
+    /**
+     * Describe the total discount price for a product (based on the quantity)
+     */
+    final double totalDiscountPrice
```
sven1103, line 34: The commercial term is `quantity discount`, and should be the financial volume difference to the total price.
```suggestion
    final double quantityDiscount
```

src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
```diff
@@ -23,10 +23,39 @@ class ProductItem {
...
+     * @param product The product for which an item is created
+     *
+     * @deprecated 2.11.0
+     */
+    @Deprecated
     ProductItem(double quantity, Product product) {
         this.quantity = quantity
         this.product = product
+        this.totalPrice = 0
+        this.totalDiscountPrice = 0
```
sven1103, line 48: renaming suggestion as above

src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
```diff
@@ -23,10 +23,39 @@ class ProductItem {
...
     }
 
+    ProductItem(double quantity, Product product, double totalPrice, double totalDiscountPrice) {
+        this.quantity = quantity
+        this.product = product
+        this.totalPrice = totalPrice
+        this.totalDiscountPrice = totalDiscountPrice
```
sven1103, line 55: same :)

sven1103 (changes requested):
Thanks a lot @jenniferboedker, I made a small comment to apply the correct financial term

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
```diff
@@ -132,6 +136,7 @@ class Offer {
          */
         double totalPrice
         double netPrice
+        double totalDiscountPrice
```
KochTobi, line 139: the official term would be `amount` for each price. So here I would suggest `discountAmount`
https://www.invoiceberry.com/free-pdf-invoice-templates

src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
jenniferboedker: Can we do quanityDiscountPrice? So its clear its a price and not the ratio? @sven1103

src/main/groovy/life/qbic/datamodel/dtos/business/Offer.groovy
jenniferboedker: But then totalPrice would also not be correct, right

sven1103: What about `appliedTotalQuantityDiscount`?

jenniferboedker: I would stay with the naming scheme we already have. At least for me its clear what it means. If we want to stick to a naming schema (not include the price) we should adapt this for the other variables as well (totalPrice, netPrice,..)

sven1103 (approved):

PR #249, opened 2021-07-29
validate ProjectSpace names

wow-such-code:
Needed to fix https://github.com/qbicsoftware/offer-manager-2-portlet/issues/695

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
@@ -23,7 +23,10 @@ final class ProjectSpace {
      */
     final String name
 
-    private static final def REGEX = ~'[-\\w]+'
+    /**
+     * Alphanumerical characters, underscores and '-' allowed
+     */
+    private static final def SPACE_NAME_REGEX = ~'[-\\w]+'
 
     ProjectSpace(String name) {
         final def space = Objects.requireNonNull(name, "Space name must not be null.")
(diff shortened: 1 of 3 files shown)


wow-such-code:
[#695 Fix: check for umlauts in space name before project creation](https://trello.com/c/94Z0gEm9/906-695-fix-check-for-umlauts-in-space-name-before-project-creation)

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
```diff
@@ -23,11 +23,21 @@ final class ProjectSpace {
      */
     final String name
 
+    private static final def REGEX = ~'[-\\w]+'
```
sven1103, line 26: I don't understand this regular expression. What does it match?

sven1103: Maybe a more declarative variable name would improve readability as well

sven1103 (changes requested):
Some minor comments only, else good!

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
wow-such-code: + matches the previous token between one and unlimited times, as many times as possible, giving back as needed (greedy)
- matches the character - with index 4510 (2D16 or 558) literally (case sensitive)
\w matches any word character (equivalent to [a-zA-Z0-9_])

not sure about the tilde, but that I took from the project code implementation

wow-such-code: pushed changes

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectSpace.groovy
```diff
@@ -23,7 +23,10 @@ final class ProjectSpace {
      */
     final String name
 
-    private static final def REGEX = ~'[-\\w]+'
+    /**
+     * Alphanumerical characters, underscores and '-' allowed
```
sven1103, line 27:
```suggestion
     * Alphanumerical characters, underscores and '-' allowed, German umlauts will lead to a mismatch
```

sven1103 (approved):
Nice!

PR #254, opened 2021-09-02
Extract project code validation

KochTobi:
This extracts the validation logic for project codes


src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectCodeValidator.groovy
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * <b>Project Code Validator</b>
+ *
+ * <p>Consumes a project code and throws a {@link ValidationException} in case it is not valid.</p>
+ *
+ * @since 2.12.0
+ */
+class ProjectCodeValidator implements Consumer<String> {
+    private static final Predicate<String> IS_VALID_PROJECT_CODE = Pattern.compile("Q[A-X0-9]{4}").asPredicate()
```
sven1103, line 17: Small extensions, to match really single project codes, and not sample codes, entities or even text with a project code contained. A project code starts with a `Q` and ends after four additional alpha-numeric letters.

```suggestion
    private static final Predicate<String> IS_VALID_PROJECT_CODE = Pattern.compile("^Q[A-X0-9]{4}$").asPredicate()
```

sven1103 (changes requested):
I suggest to move the validation class now out of the DTO package.

+1 for lambda support via the functional interfaces.

sven1103:
Ah and the CL needs an update as well :)

src/main/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectCodeValidator.groovy
KochTobi: Nice catch 
image

sven1103 (approved):

PR #255, opened 2021-09-10
Shorthand label for Facility enum

sven1103:
This PR provides a custom shorthand label for the Facility enum class.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/facilities/Facility.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/facilities/Facility.groovy
@@ -27,21 +27,12 @@ enum Facility {
     private final String label
 
     /**
-     * Creates an instance of a facility enum
-     * @param fullName The full name representation of the facility
-     */
-    Facility(String fullName) {
-        this.fullName = fullName
-        this.label = ""
-    }
-
-    /**
      * Creates an instance of a facility enum with shorthand label
      * @param fullName The full name representation of the facility
      * @param label The shorthand label of the facility
      */
     Facility(String fullName, String label) {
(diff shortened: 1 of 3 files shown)


src/main/groovy/life/qbic/datamodel/dtos/business/facilities/Facility.groovy
```diff
@@ -17,22 +17,35 @@ package life.qbic.datamodel.dtos.business.facilities
...
 
     /**
      * Creates an instance of a facility enum
-     * @param fullName The full name representation of the enum
+     * @param fullName The full name representation of the facility
      */
     Facility(String fullName) {
         this.fullName = fullName
+        this.label = ""
     }
```
KochTobi, line 36: Why do you provide this constructor. In case a label is always expected (seems like it from the current emum objects) this constructor will not be used and can be removed.

src/main/groovy/life/qbic/datamodel/dtos/business/facilities/Facility.groovy
```diff
@@ -17,22 +17,35 @@ package life.qbic.datamodel.dtos.business.facilities
...
 
+    /**
+     * Creates an instance of a facility enum with shorthand label
+     * @param fullName The full name representation of the facility
+     * @param label The shorthand label of the facility
+     */
+    Facility(String fullName, String label) {
+        this(fullName)
```
KochTobi, line 44: You can replace this constructor chaining by removing the old one and integrating the one line here instead
```suggestion
        this.fullName = fullName
```

KochTobi (changes requested):

KochTobi (approved):

PR #258, opened 2021-09-13
Hotfix 2.12.1

sven1103:
Introduces an important fix that caused an NPE when comparing `life.qbic.datamodel.dtos.projectmanagement.ProjectIdentifier` to a null reference.

The expected return value is `false`, which is now introduced.

### DIFF
--- a/src/test/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifierSpec.groovy
+++ b/src/test/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifierSpec.groovy
@@ -40,6 +40,10 @@ class ProjectIdentifierSpec extends Specification{
 
     }
 
+    /**
+     * Tests for Joshua Bloch's Item 8 in Effective Java, how to override equals() properly.
+     * https://biratkirat.medium.com/learning-effective-java-item-8-d05f3847213d
+     */
     def "Test equals method fulfills full method contract"() {
         when:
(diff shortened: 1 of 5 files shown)


src/test/groovy/life/qbic/datamodel/dtos/projectmanagement/ProjectIdentifierSpec.groovy
```diff
@@ -40,4 +40,22 @@ class ProjectIdentifierSpec extends Specification{
...
+        (idX.equals(idY) == true) && (idY.equals(idX) == true) // Symmetric
+        (idX.equals(idY) == true) && (idY.equals(idZ) == true) && (idX.equals(idZ) == true) // Transitive
+        (idX.equals(idX) == true) && (idX.equals(idX) == true) // Consistent
+        idX.equals(null) == false // Non-nullity
+
+        where:
+        projectSpace | projectCode
+        "TEST" | "QTEST"
+    }
+
```
KochTobi, line 60: I think it would be better to split this into three tests for all the different properties.

sven1103: why would it be better?

KochTobi: It would make the tests simpler. You wouldn't need to look into the code and read the comments to understand what is tested here.

KochTobi (approved):

PR #260, opened 2021-09-14
add updated DS figure

jenniferboedker:
Updates the current maxquant file structure


Steffengreiner (approved):
Nice 🚀 
However why do we have a SNAPSHOT version on the master branch? (This is the reason why qube linting fails)

PR #288, opened 2022-02-15
Remove everit-json from compile scope

KochTobi:
Removes everit.json from compile scope so that the `jitpack` repository does not need to be imported by users of this lib.
```
<groupId>com.github.everit-org.json-schema</groupId>
<artifactId>org.everit.json.schema</artifactId>
```
Further, this PR simplifies metadata format validation and removes the specification for it in form of a JSON schema (as it is straightforward and property names are hardcoded anyhow). Validation is still checked in the test specifications.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
@@ -66,7 +66,6 @@ final class OxfordNanoporeMeasurement {
 
     private static void validateMetaData(Map metadata) throws IllegalArgumentException {
         def expectedKeys = [
-                "adapter",
                 "asic_temp",
                 "base_caller",
                 "base_caller_version",
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
@@ -67,8 +67,8 @@ final class OxfordNanoporeMeasurement {
 
     private static void validateMetaData(Map metadata) throws IllegalArgumentException {
 
-        InputStream stream = Thread.currentThread().getContextClassLoader().getResourceAsStream("schemas/ont-metadata.schema.json")
-        Map schema = (Map) new JsonSlurper().parse(stream)
+        InputStream stream = this.getClassLoader().getResourceAsStream("schemas/ont-metadata.schema.json")
+        def schema = new JsonSlurper().parse(stream) as Map<String, ?>
 
         def expectedKeys = schema.get("required") as List<String>
 
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
@@ -1,5 +1,6 @@
 package life.qbic.datamodel.datasets
 
+import groovy.json.JsonSlurper
 import groovy.util.logging.Log4j2
 import life.qbic.datamodel.datasets.datastructure.files.DataFile
 import life.qbic.datamodel.datasets.datastructure.folders.DataFolder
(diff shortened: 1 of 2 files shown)


src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -368,25 +374,4 @@ final class OxfordNanoporeMeasurement {
...
-            final def schemaStream = OxfordNanoporeMeasurement.getResourceAsStream(SCHEMA)
-            final def rawSchema = new JSONObject(new JSONTokener(schemaStream))
-            final def jsonSchema = SchemaLoader.load(rawSchema)
-            // Validate against schema
-            jsonSchema.validate(metaDataJson)
-        }
-
-    }
-
-
```
jenniferboedker, line 391: So this means we cannot make use of the schemata we defined?

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -70,12 +65,23 @@ final class OxfordNanoporeMeasurement {
...
+                "base_caller",
+                "base_caller_version",
+                "device_type",
+                "flow_cell_id",
+                "flow_cell_position",
+                "flow_cell_product_code",
+                "protocol",
+                "hostname",
+                "started"
+        ]
```
jenniferboedker, line 79: This looks redundant to me, it's actually the same as defined in the METADATA_FIELD. What about using the enum and defining these strings there too. Then you can say the expected keys are the values defined in the enum METADATA_FIELD

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -368,25 +374,4 @@ final class OxfordNanoporeMeasurement {
     String getRelativePath() {
         return this.measurementFolder.relativePath
     }
-
-    /*
-    Inner class that contains the logic for the metadata validation
-     */
-    private static class MetaData {
-
-        private static final SCHEMA = "/schemas/ont-metadata.schema.json"
```
jenniferboedker, line 377: Since we defined the schema I think it would make sense to use it. How about using the JSON Slurper?

jenniferboedker (changes requested):
I left some comments. I think it would make sense to still use the schema we have defined. Otherwise, we get out of sync.

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
KochTobi: Yes we can but in the core-utils lib where the parsers reside and not in the data-model-lib

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
KochTobi: I did not manage to validate a schema with JSON Slurper. I won't address that. Please feel free to make a suggestion

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
KochTobi: This is used to separate the way the information is provided (we can not influence that) from the way we process the metadata. Not all metadata that will be processed is required (adapter for example is not) so this needs to be a separate list.

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
jenniferboedker: I pushed my suggestion as discussed, this way we use the schema and do not need to hardcode the keys

jenniferboedker (dismissed):

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -369,24 +312,122 @@ final class OxfordNanoporeMeasurement {
...
+        private String asicTemp
+        private String deviceType
+        private String flowcellId
+        private String flowcellPosition
+        private String flowcellType
+        private String baseCaller
+        private String baseCallerVersion
+        private String libraryPreparationKit
+        private String machineHost
+        private String startDate
```
jenniferboedker, line 329: can we make them public but final? so we dont need the getter and just provide immutable content?

jenniferboedker: You will always use the same schema and parse it with the constructor

jenniferboedker (approved):

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
KochTobi: This way we do not have to introduce a new constructor for every eventuality but can use the factory method. I'd say this is good as is.

PR #289, opened 2022-02-17
add alternative nanopore data structure and tests

wow-such-code:
* mux_scan_data_*.csv and drift_correction_*.csv can alternatively be situated in the new folder "other_reports"

* barcode_alignment_*.tsv is a new file that is now recognized


src/main/resources/schemas/nanopore-instrument-output.schema.json
```diff
@@ -407,6 +438,9 @@
                   },
                   {
                     "$ref": "#/definitions/throughput_log"
+                  },
```
wow-such-code, line 441: please note, that "minItems" does not change: we add 1 new file and allow 2 old files to move to a new folder

src/test/resources/valid-example-newer.json
```diff
@@ -0,0 +1,200 @@
...
+                            "name": "myfile.fast5",
+                            "path": "./20200122_1217_1-A1-B1-PAE12345_1234567a/fast5_pass/myfile.fast5",
+                            "file_type": "fast5"
+                        }
+                    ]
+                }
+            ]
+        }
+    ]
+}
```
wow-such-code, line 200:
```suggestion
}

```

jenniferboedker (commented):
Looks good so far and makes sense, you might also need to update the datamodel description (in the README) it looks like the old and the new structure both need to be accepted, right?

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
```diff
@@ -112,6 +112,8 @@ final class OxfordNanoporeMeasurement {
     }
 
     private static String extractLibraryKit(String text) {
+        // cut off optional suffix
+        text = text.replace(":True", "")
```
jenniferboedker, line 116: Where does the suffix come from? And why cut it off, would it be required maybe for the labs? This would then change the file name, right?

src/main/groovy/life/qbic/datamodel/datasets/datastructure/files/nanopore/BarcodeAlignmentLog.groovy
```diff
@@ -0,0 +1,33 @@
+package life.qbic.datamodel.datasets.datastructure.files.nanopore
+
+import life.qbic.datamodel.datasets.datastructure.files.DataFile
+
+/**
+ * A specialisation of a DataFile, represents an Oxford Nanopore barcode alignment log file
+ *
+ * @author: Sven Fillinger
```
jenniferboedker, line 8:
```suggestion
```

jenniferboedker (changes requested):
With all the changes there are I think it would be better to separate the schemata. 

We should have two of them (the second could be ont-v2) since these are two versions of a machine output of the same technology. We would then need a second experiment type respectively. And this would also mean we would need a second parser in CUL. Maybe we can quickly discuss this in person?

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeMeasurement.groovy
wow-such-code: we are not using it for anything and it was cut off before, just using a regex. now we have the cases where it is missing and where we need to cut it off, because the regex is not working for the version where it is already missing.

src/main/resources/schemas/nanopore-instrument-output_v2.schema.json
```diff
@@ -0,0 +1,605 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema",
+  "$id": "http://qbic.life/nanopore-instrument-output_v2.schema.json",
+  "title": "Nanopore Instrument Output",
```
jenniferboedker, line 4:
```suggestion
  "title": "Nanopore Instrument Output V2",
```

src/main/resources/schemas/nanopore-instrument-output.schema.json
```diff
@@ -390,6 +390,9 @@
                   {
                     "$ref": "#/definitions/duty_time_log"
                   },
+                  {
+                    "$ref": "#/definitions/barcode_alignment_log"
+                  },
```
jenniferboedker, line 395: Are you sure this will be needed in the old schema as well? Mibi never communicated this requirement to us

src/main/groovy/life/qbic/datamodel/instruments/OxfordNanoporeInstrumentOutputV2.groovy
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @author Sven Fillinger
+ * @since 1.9.0
+ */
+class OxfordNanoporeInstrumentOutputV2 {
+
+    private static final String SCHEMA_PATH = "/schemas/nanopore-instrument-output_v2.schema.json"
+
+    static InputStream getSchemaAsStream() {
+        return OxfordNanoporeInstrumentOutput.getResourceAsStream(SCHEMA_PATH)
```
jenniferboedker, line 20:
```suggestion
        return OxfordNanoporeInstrumentOutputV2.getResourceAsStream(SCHEMA_PATH)
```

src/main/groovy/life/qbic/datamodel/datasets/datastructure/folders/nanopore/OtherReportsFolder.groovy
```diff
@@ -0,0 +1,41 @@
+package life.qbic.datamodel.datasets.datastructure.folders.nanopore
+
+import life.qbic.datamodel.datasets.datastructure.folders.DataFolder
+
+/**
+ * A special case of a DataFolder.
+ *
+ * Holds information about a Oxford Nanopore NGS measurement.
+ *
+ * @author: Sven Fillinger
```
jenniferboedker, line 10:
```suggestion
```

README.md
```diff
@@ -101,7 +101,7 @@ Make sure, that you have defined the Github package Maven repository, in order f
 
 A Nanopore NGS measurement output is delivered to us as a nested folder structure, following this model:
 
-![Nanopore Data Structure Model](./doc/figures/Nanopore_Data_Structure_Model.png)
+![Nanopore Data Structure Model](./doc/figures/Nanopore_Data_Structure_Model.svg)
```
jenniferboedker, line 104: NIT: if we have two schemas we need to figures

jenniferboedker (commented):
Thanks for taking my comments into consideration :) 
Only a few more comments, after you addressed them we are good to go 🚀

src/main/resources/schemas/nanopore-instrument-output.schema.json
wow-such-code: I agree, I will remove it

README.md
wow-such-code: solved

jenniferboedker (approved):
I love it, thanks for addressing even my NIT! 💯

PR #295, opened 2022-02-25
update groovy versions

KochTobi:
Upgrades the Groovy version to `3.0.9`
Upgrades to Java 11 (needed for xml-manager-lib with Java 11)


wow-such-code (approved):
lgtm

PR #317, opened 2022-05-11
update maxquant schema

wow-such-code:
Updates the MaxQuant model for incoming data: experimentalDesignTemplate, summary and an unnecessary folder are removed. The sample ids file now needs a valid QBiC project code.


src/main/groovy/life/qbic/datamodel/datasets/MaxQuantRunResult.groovy
```diff
@@ -54,23 +50,19 @@ final class MaxQuantRunResult {
 
     private final SampleIds sampleIds
 
-    private final Summary summary
-
-    MaxQuantRunResult(AllPeptides allPeptides, Evidence evidence, ExperimentalDesignTemplate experimentalDesignTemplate, Parameters parameters, Peptides peptides, ProteinGroups proteinGroups, RunParameters runParameters, SampleIds sampleIds, Summary summary) {
+    MaxQuantRunResult(AllPeptides allPeptides, Evidence evidence, Parameters parameters, Peptides peptides, ProteinGroups proteinGroups, RunParameters runParameters, SampleIds sampleIds) {
```
jenniferboedker, line 53: If we remove those values from the constructor we need to either create a new major release now or make the fields and constructor deprecated and then create a new constructor so that both are available.

jenniferboedker (changes requested):
Thanks for tackeling this task:)
Looks good for me. Can you maybe elaborate on how you planned to release this lib. AFAIK its not good to just remove fields and change constructors. I would rather prefer to have either deprecated methods or you need to make sure a new major release is created since this changes introduce "breaking changes" (see my comment in the code)

src/main/groovy/life/qbic/datamodel/datasets/MaxQuantRunResult.groovy
wow-such-code: agree, I changed it!

jenniferboedker (approved):
looks good to me now :)
left one nit comment though

PR #335, opened 2022-10-21
Add positional information for offer product items

sven1103:
Enables the client to store positional information of product items.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
+++ b/src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
@@ -41,6 +41,13 @@ class ProductItem {
     final double quantityDiscount
 
     /**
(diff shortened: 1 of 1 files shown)


src/main/groovy/life/qbic/datamodel/dtos/business/ProductItem.groovy
```diff
@@ -40,6 +40,13 @@ class ProductItem {
      */
     final double quantityDiscount
 
+    /**
+     * Stores the latest position on the offer.
+     *
+     * A negative value indicates no specific position was stored in the item.
+     */
+    private int offerPosition = -1
```
jenniferboedker, line 48: why private? the other fields are not set private.

I would do it consistent for all fields. Either private setters or not. And I think there was a reason to not use them but I cant remember

jenniferboedker (approved):

PR #338, opened 2022-11-08
Introduce reference Id in persons

sven1103:


src/main/groovy/life/qbic/datamodel/dtos/general/Person.groovy
```diff
@@ -27,6 +27,12 @@ abstract class Person {
    */
   final int id
 
+  /**
+   * Unique entity identifier (UUID)
+   * @since 2.22.0
+   */
+  final String referenceId
```
KochTobi, line 34: I think we should store it as UUID here and only convert for the database

sven1103: why?

KochTobi: Ah nvm those are only the DTOs

KochTobi (approved):

PR #340, opened 2022-11-16
Introduce inactive flag for affiliations

sven1103:
Allows for filtering active affiliations


KochTobi (approved):
What is the name of the getter generated by groovy `isActive()` or `getActive()`

sven1103:
If I remember correctly, boolean values will result in `is<Field name>`

PR #341, opened 2022-11-23
Feature/dm 589 allow fractions in all products

Steffengreiner:
Adapts all service products to use partial products instead of atomic product since fractional input should be allowed 
**Question**
 Should this PR also remove the AtomicProduct definition since it's not used anymore?


wow-such-code (approved):
lgtm

PR #350, opened 2023-01-31
Introduce minimal Schema for nanopore registration 

Steffengreiner:
**What was changed**
This PR introduces a schema and example json for checking if the minimal required files for a nanopore data registration are within a provided dataset. 
Additionally it adapts the `OxfordNanoporeExperiment` generation to not fail if an unknown file or folder is found but include them as `OptionalFile` and `OptionalFolder` in the datastructure to allow for variations within the provided datastructure.

**Background information**
As of now the minimum required folders are: 

- FASTQ_FAIL
- FASTQ_PASS
- FAST5_FAIL
- FAST5_PASS

While the minimum required files are: 

- final_summary.txt (We extract metadata from this file)
- report.md (We extract metadata from this file)
- sequencing_summary.txt (Both Labs)

**Open Questions**
Due to the new constraints, the previously defined schemas and datafiles/Datafolders are ALL now unnecessary and could be deleted, since as soon as the minimal required files are provided the `OxfordNanoporeExperiment` is created and can be parsed.
The only reason for keeping multiple schemas could be if the distinction between the files in the OxfordnanoporeExperiment class is necessary (e.g. for accessing metadata within the files).
However this is only the case for the `final_summary.txt` file and the `report.md` file as of yet.

### DIFF
--- a/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
+++ b/src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
@@ -16,54 +16,9 @@ import java.lang.reflect.Method
  */
 final class OxfordNanoporeExperiment implements ExperimentFolder {
 
-  // Fully qualified domain name of the nanopore folder structure package
-  private static final String FQDN_FOLDERS = "life.qbic.datamodel.datasets.datastructure.folders.nanopore"
-  // Fully qualified domain name of the nanopore file structure package
-  private static final String FQDN_FILES = "life.qbic.datamodel.datasets.datastructure.files.nanopore"
-
-  //Unique key in fileTreeMap identifying a child as a file
-  private static final String FILEKEY = "file_type"
-  //Unique key in fileTreeMap identifyng a child as a folder
-  private static final String FOLDERKEY = "children"
-
   private final List<OxfordNanoporeMeasurement> measurements
 
   private final String sampleId
-
-  private final static Set nanoporeFileTypes = [
-          FQDN_FILES + ".DriftCorrectionLog",
-          FQDN_FILES + ".DutyTimeLog",
-          FQDN_FILES + ".Fast5File",
-          FQDN_FILES + ".FastQFile",
-          FQDN_FILES + ".FastQZippedFile",
-          FQDN_FILES + ".FinalSummaryLog",
-          FQDN_FILES + ".MuxScanDataLog",
-          FQDN_FILES + ".ReportMdLog",
-          FQDN_FILES + ".ReportPDFLog",
-          FQDN_FILES + ".ReportHTMLLog",
-          FQDN_FILES + ".ReportJSONLog",
-          FQDN_FILES + ".SequencingSummaryLog",
-          FQDN_FILES + ".ThroughputLog",
-          FQDN_FILES + ".BarcodeAlignmentLog",
-          FQDN_FILES + ".PoreActivityLog",
-          FQDN_FILES + ".SampleSheetLog",
-          FQDN_FILES + ".PoreScanDataLog",
-          FQDN_FILES + ".SequencingTelemetryLog",
-          FQDN_FILES + ".GuppyBasecallLog"
-  ]
-
-  private final static Set nanoporeFolderTypes = [
-          FQDN_FOLDERS + ".Fast5Folder",
-          FQDN_FOLDERS + ".FastQFolder",
-          FQDN_FOLDERS + ".Fast5PassFolder",
-          FQDN_FOLDERS + ".Fast5FailFolder",
-          FQDN_FOLDERS + ".FastQPassFolder",
-          FQDN_FOLDERS + ".FastQFailFolder",
-          FQDN_FOLDERS + ".UnclassifiedFast5Folder",
-          FQDN_FOLDERS + ".UnclassifiedFastQFolder",
-          FQDN_FOLDERS + ".OtherReportsFolder",
-          FQDN_FOLDERS + ".BasecallingFolder"
-  ]
 
   private OxfordNanoporeExperiment(String sampleId, List<OxfordNanoporeMeasurement> measurements) {
     this.measurements = Objects.requireNonNull(measurements, "measurements must not be null")
@@ -96,8 +51,7 @@ final class OxfordNanoporeExperiment implements ExperimentFolder {
   }
 
   /**
-   * Helper method that parses the QBiC identifier from the root folder name
-   */
+   * Helper method that parses the QBiC identifier from the root folder name*/
   private static String parseQbicIdFromRootFolder(Map nanoPoreSequencerOutput) {
     def name = Objects.requireNonNull(nanoPoreSequencerOutput.get("name"), "The root folder must contain a name property.")
     final def ids = SampleCodeFunctions.findAllQbicSampleCodes(name as String)
@@ -111,8 +65,7 @@ final class OxfordNanoporeExperiment implements ExperimentFolder {
   }
 
   /**
-   * Helper method that creates the measurements from the sequencer output
-   */
+   * Helper method that creates the measurements from the sequencer output*/
   private static List<OxfordNanoporeMeasurement> parseMeasurements(Map nanoPoreSequencerOutput) {
     final def measurements = []
     Objects.requireNonNull(nanoPoreSequencerOutput.get("children"), "The root folder must contain at least one measurement folder.")
@@ -134,15 +87,17 @@ final class OxfordNanoporeExperiment implements ExperimentFolder {
     final def children = []
     items.each { item ->
       {
-        if (item.get(FILEKEY)) {
-          // Lets try to parse it as a subclass of a DataFile
-          DataFile putativeFile = parseFile(item)
-          children.add(putativeFile)
-        }
-        if (item.get(FOLDERKEY)) {
-          // Lets try to parse it as a subclass of a DataFile
-          DataFolder putativeFolder = parseFolder(item)
-          children.add(putativeFolder)
+        if (!item.isEmpty()) {
+          if (isFile(item)) {
+            // Lets try to parse it as a subclass of a DataFile
+            DataFile putativeFile = parseFile(item)
+            children.add(putativeFile)
+          }
+          if (isFolder(item)) {
+            // Lets try to parse it as a subclass of a DataFile
+            DataFolder putativeFolder = parseFolder(item)
+            children.add(putativeFolder)
+          }
         }
       }
     }
@@ -156,7 +111,7 @@ final class OxfordNanoporeExperiment implements ExperimentFolder {
   private static DataFile parseFile(Map fileTree) throws IllegalArgumentException {
     String name = fileTree.get("name")
     String path = fileTree.get("path")
-    for (String nanoPoreFileType : nanoporeFileTypes) {
+    for (String nanoPoreFileType : NanoporeFileTypes.values()) {
       Class<?> c = Class.forName(nanoPoreFileType)
       Method method = c.getDeclaredMethod("create", String.class, String.class)
       try {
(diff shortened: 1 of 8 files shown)


src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-            throw new IllegalArgumentException("No QBiC sample identifier found!")
+  /*
+   * Helper method that creates a list of mixed DataFolders and DataFiles instances
+   */
+
+  private static List<?> parseMeasurementItems(List<Map> items) {
+    final def children = []
+    items.each { item ->
+      {
+        if (item.get(FILEKEY)) {
```
KochTobi, line 137: Maybe extract this into a method `isFile():boolean` and make the null or empty check more expressive. It took me a while to understand that you check whether the item has a property called `file_type` provided by `FILEKEY` you check for here.

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
+          // Lets try to parse it as a subclass of a DataFile
+          DataFile putativeFile = parseFile(item)
+          children.add(putativeFile)
         }
-        if (ids.size() > 1) {
-            throw new IllegalArgumentException("Name contained more than one valid sample id!")
+        if (item.get(FOLDERKEY)) {
```
KochTobi, line 142: Same as with the `FILEKEY`, I think this is not very easy to read and understand.

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-        final def ids = SampleCodeFunctions.findAllQbicSampleCodes(name as String)
-        if (ids.isEmpty()) {
-            throw new IllegalArgumentException("No QBiC sample identifier found!")
+  /*
+   * Helper method that creates a list of mixed DataFolders and DataFiles instances
+   */
+
+  private static List<?> parseMeasurementItems(List<Map> items) {
+    final def children = []
+    items.each { item ->
```
KochTobi, line 135: Why did you change from `IllegalArgumentException` to the differentiation by properties present?

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
+    def parsedChildren = []
+    children.each { Map unknownChild ->
+      if (unknownChild.get(FILEKEY)) {
+        def child = parseFile(unknownChild)
+        parsedChildren.add(child)
+      }
+      if (unknownChild.get(FOLDERKEY)) {
+        // We do not capture the second parse call, as we want to fail the parsing at this point.
+        // This means that we ultimately found a child of unknown type, which should
+        // break the parsing.
```
KochTobi, line 250: Is this still valid?

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
+        def child = parseFolder(unknownChild)
+        parsedChildren.add(child)
+      }
```
KochTobi, line 253: reminds me of the measurement item parsing. Please refer to my comments there.

KochTobi (changes requested):
Only some minor requests

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-            FQDN_FOLDERS + ".OtherReportsFolder",
-            FQDN_FOLDERS + ".BasecallingFolder"
-    ]
-
-    private OxfordNanoporeExperiment(String sampleId, List<OxfordNanoporeMeasurement> measurements) {
-        this.measurements = Objects.requireNonNull(measurements, "measurements must not be null")
-        this.sampleId = Objects.requireNonNull(sampleId, "sampleId must not be null")
-    }
+  // Fully qualified domain name of the nanopore folder structure package
+  private final static String FQDN_FOLDERS = "life.qbic.datamodel.datasets.datastructure.folders.nanopore"
```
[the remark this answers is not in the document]
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-     * @param Map nanoPoreSequencerOutput
-     * @return OxfordNanoporeExperiment A new instance of a nanopore experiment.
-     */
-    static OxfordNanoporeExperiment create(Map nanoPoreSequencerOutput) {
-        final String sampleId = parseQbicIdFromRootFolder(nanoPoreSequencerOutput)
-        final List<OxfordNanoporeMeasurement> measurements = parseMeasurements(nanoPoreSequencerOutput)
-        return new OxfordNanoporeExperiment(sampleId, measurements)
-    }
+  //Unique key in fileTreeMap identifying a child as a file
+  private final static String FILEKEY = "file_type"
```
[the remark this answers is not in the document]
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
+  //Unique key in fileTreeMap identifyng a child as a folder
+  private final static String FOLDERKEY = "children"
+
+  private final List<OxfordNanoporeMeasurement> measurements
+
+  private final String sampleId
+
+  private final static Set nanoporeFileTypes = [
```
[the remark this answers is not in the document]
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-     */
-    static OxfordNanoporeExperiment create(Map nanoPoreSequencerOutput) {
-        final String sampleId = parseQbicIdFromRootFolder(nanoPoreSequencerOutput)
-        final List<OxfordNanoporeMeasurement> measurements = parseMeasurements(nanoPoreSequencerOutput)
-        return new OxfordNanoporeExperiment(sampleId, measurements)
-    }
+  //Unique key in fileTreeMap identifying a child as a file
+  private final static String FILEKEY = "file_type"
+  //Unique key in fileTreeMap identifyng a child as a folder
+  private final static String FOLDERKEY = "children"
```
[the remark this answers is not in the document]
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
-    ]
-
-    private OxfordNanoporeExperiment(String sampleId, List<OxfordNanoporeMeasurement> measurements) {
-        this.measurements = Objects.requireNonNull(measurements, "measurements must not be null")
-        this.sampleId = Objects.requireNonNull(sampleId, "sampleId must not be null")
-    }
+  // Fully qualified domain name of the nanopore folder structure package
+  private final static String FQDN_FOLDERS = "life.qbic.datamodel.datasets.datastructure.folders.nanopore"
+  // Fully qualified domain name of the nanopore file structure package
+  private final static String FQDN_FILES = "life.qbic.datamodel.datasets.datastructure.files.nanopore"
```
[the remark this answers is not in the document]
Steffengreiner: Nice catch 👀

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
```diff
@@ -14,239 +16,258 @@ import java.lang.reflect.Method
...
+          FQDN_FILES + ".ThroughputLog",
+          FQDN_FILES + ".BarcodeAlignmentLog",
+          FQDN_FILES + ".PoreActivityLog",
+          FQDN_FILES + ".SampleSheetLog",
+          FQDN_FILES + ".PoreScanDataLog",
+          FQDN_FILES + ".SequencingTelemetryLog",
+          FQDN_FILES + ".GuppyBasecallLog"
+  ]
+
+  private final static Set nanoporeFolderTypes = [
```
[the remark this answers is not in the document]
Steffengreiner: Good idea 🧠

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: Good idea 🧠

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: It's adapted nice idea 👍

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: Will do 🥳

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: Basically to avoid unnecessary exception generation since we now know what defines a file/folder before iterating over all possible combinations and allow for unknown items in the map. 

**In more detail:** 
Since we now allow for optional files and folders, parseFile() and parseFolder() will not throw an exception if the item is not recognized. Instead it will create an OptionalFile with the content of the provided item. 
However this is not possible if the item is a folder, since a folder does not have a file_type, which is required for the optional file generation. This leads to an exception which would have to be caught and redirected again. 

Also IMO It's more explicit since we explicitly show what the defining properties are since we rely on them during the File/Folder generation process anyway.

KochTobi: The original reasoning was that what makes a file a file and what makes a folder a folder is defined in the JSON schema. Thus only the method that uses the schema can determine whether we are indeed parsing a file or a folder. It first checks whether a file can be parsed, if not it tries to parse a folder. If neither works, then there should be an exception. The solution suggested in this PR does not consider that and instead hard-codes the properties.

KochTobi: As we have this logic in multiple places, perhaps it would be an option to extract the parsing or to restructure the code slightly.
If this is more complex than suspected or if I missed some aspects of it, I will not block this PR as it works.

KochTobi (approved):
LGTM :+1:

src/main/groovy/life/qbic/datamodel/datasets/OxfordNanoporeExperiment.groovy
Steffengreiner: I'd agree with you however the parseFile and parseFolder methods beforehand also expected specific properties to be there in the first place (namely "name", "path", and "children"). 
Additionally the parsing should not happen if the schema fails, therefore we can expect that these properties could be found during the validation process.

Steffengreiner: I'd argue that down the line we should re-evaluate where the schema should be processed. 
IMO the strict seperation of schema and parsing does not make sense anymore and the schema should be automatically consulted if a NanoporeExperiment/NanoporeMeasurement is generated. 🤷‍♂️

PR #368, opened 2023-11-02
Add support for pod5 files generated via dorado basecaller

Steffengreiner:
**What was changed**
Apparently the new Nanopore machines generate pod5 files instead of fast5 files dependent on the employed basecaller. Since the partner labs are updating their machines it's necessary to account for the newly valid minimal filetree structure

PR #369, opened 2023-11-03
Release 0.27.0

Steffengreiner:
Add support for pod5 files generated via dorado basecaller #368


Steffengreiner:
Main and Dev diverged so it has to be fixxed in a seperate branch 😢

PR #374, opened 2023-11-09
Allow Nf-Core dataset registration without providing a runId file

Steffengreiner:
**What was changed**
Adaption of the datastructure to allow the runId to be nullable since it was provided only on nf-tower runs.


src/main/groovy/life/qbic/datamodel/datasets/NfCorePipelineResult.groovy
```diff
@@ -67,6 +68,17 @@ final class NfCorePipelineResult {
         this.sampleIds = sampleIds
     }
 
+    NfCorePipelineResult(PipelineInformationFolder pipelineInformationFolder, QualityControlFolder qualityControlFolder, List<DataFolder> processFolders, SampleIds sampleIds) {
+        Objects.requireNonNull(pipelineInformationFolder, "Please provide a PipelineInformation folder.")
+        Objects.requireNonNull(qualityControlFolder, "Please provide a QualityControl folder")
+        Objects.requireNonNull(processFolders, "Please provide a List of process folders")
+        Objects.requireNonNull(sampleIds, "Please provide a sampleIds file")
```
wow-such-code, line 75: why is sample ids nullable in one of the constructors?

wow-such-code (changes requested):
a tiny NIT, which might improve the code

wow-such-code (approved):
never mind, nice work

